PR #24, opened 2021-01-16
Create package structure

rvighne:
Initialized an NPM package for the app and created stubs of client and server programs. All 4 programs (controller, client, daemon, frontend) will live in the same package so they can share dependencies and source code.

- `server.js` is the entrypoint for the daemon. It should be short and just trigger other code.
- `client.js` is the entrypoint for the controller or client. It should be short and just trigger other code.
- `src` will contain our classes. It will be shared among all 4 programs so you can just import any preexisting code.

Not included here, but will be needed:

- Unit testing framework (I did some research and would recommend [Jest](https://jestjs.io/)).
- CI/CD support (Travis CI).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+node_modules
--- a/client.js
+++ b/client.js
@@ -1,0 +1,2 @@
+console.error('Not implemented')
+process.exit(1)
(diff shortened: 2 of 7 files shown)


alexzma:
This looks good to me. Not sure if we should include unit testing and CI/CD before merging to master though.

rvighne:
@alexzma Yeah, I expect that can be a separate PR. This will let us start coding in the meantime.

alexzma:
That sounds good. I'll approve the merge then.

PR #29, opened 2021-01-21
Update Travis and release badge URLs

rvighne:
Yet another badge update PR... This time the badge has been tested, and the other badge (latest release version number) is also updated to point to this repo instead of the template.


alexzma (approved):
This looks good to me. Interesting that travis-ci.org still worked since they announced that they would move all projects to using travis-ci.com in 2018.

yashlala (approved):

PR #30, opened 2021-01-21
Configure Typescript, Linting, and Formatting

Rohanator9000:


tsconfig.json
```diff
@@ -0,0 +1,71 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
+    // "allowJs": true,                       /* Allow javascript files to be compiled. */
```
rvighne, line 10: Probably best to disable this since we're using TS from the beginning.

.prettierrc
```diff
@@ -0,0 +1,14 @@
+{
+    "overrides": [
+        {
+            "files": "*.ts",
+            "options": {
+            "parser": "typescript"
```
rvighne, line 6: Indentation

package.json
```diff
@@ -3,9 +3,22 @@
   "version": "1.0.0",
   "description": "Zero-config distributed parallel processing",
   "license": "Apache-2.0",
-  "main": "client.js",
+  "main": "./dist/server.js",
   "repository": "github:cs130-w21/1",
   "scripts": {
```
rvighne, line 8: We also need a script to run prettier, right?

.vscode/extensions.json
```diff
@@ -0,0 +1,13 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
```
rvighne, line 6: Can you also recommend a vscode extension that runs prettier? (unless it has that built-in)

package.json
```diff
@@ -3,9 +3,22 @@
...
+    "compile": "npx tsc",
+    "compile:watch": "npx tsc -w",
+    "lint": "npx eslint .",
+    "start": "node ./dist/server.js",
+    "start:dev": "ts-node ./src/server.ts",
     "test": "echo \"Error: no test specified\" && exit 0"
```
rvighne, line 14: Eslint and prettier should be verified by the test command.

rvighne (changes requested):

.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+dist
```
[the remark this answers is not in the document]
yashlala: They serve different functions -- "don't track this" != "don't format this". Linking unrelated tools can lead to unexpected behavior (eg. I set up a temporary playground with `git update-index --skip-worktree`, now its not going to auto-format). 

Current commit seems fine as is.

.vscode/extensions.json
yashlala: Why do we track personal editor choices globally? Is this usual practice? 

(It's hard to read tone over text, so I'll be explicit: I'm aware this issue isn't worth the attention, I'm just curious as to whether this is a common practice. I was under the impression teams usually have a strict moratorium on personal configuration files.)

rvighne: It varies. Open source projects usually don't, companies usually do. In this case it's just a suggestion rather than actual settings, and most of us will be using vscode anyway, so I'm fine with it.

.eslintignore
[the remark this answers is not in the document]
rvighne: `.gitignore` only prevents files from being added, so it's unaffected by `--skip-worktree` and you could still use that as usual. Nevertheless, might be good to have flexibility, so I'm fine either way.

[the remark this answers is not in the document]
yashlala: ik, I use skip worktree to ignore local changes to my gitignore (so I can put work in a tmpdir while keeping the output of `git status` nice. 

In any case marking as resolved.

yashlala (approved):
No additional changes required (discounting the open change requests).

Rohanator9000:
@rvighne I added ts-node because nodemon deals just with JS, meaning that any stack traces will point to JS files. With ts-node they point directly to the TS files, which I find more helpful since that's what I'm actually editing.

rvighne:
> I added ts-node because nodemon deals just with JS, meaning that any stack traces will point to JS files. With ts-node they point directly to the TS files, which I find more helpful since that's what I'm actually editing.

Ah thanks for pointing this out. This is a property of Node rather than nodemon; I need to enable source map support, and then the exceptions will refer to TS locations. Another reason for nodemon over ts-node: it doesn't add a redundant compilation step (we're already doing that with tsc) and it detects file changes so it can auto-restart the server, will be very convenient during dev.

Rohanator9000:
> > I added ts-node because nodemon deals just with JS, meaning that any stack traces will point to JS files. With ts-node they point directly to the TS files, which I find more helpful since that's what I'm actually editing.
> 
> Ah thanks for pointing this out. This is a property of Node rather than nodemon; I need to enable source map support, and then the exceptions will refer to TS locations. Another reason for nodemon over ts-node: it doesn't add a redundant compilation step (we're already doing that with tsc) and it detects file changes so it can auto-restart the server, will be very convenient during dev.

Source maps are on in `tsconfig.json`. If we can get TS errors with nodemon, I'm all for using it. Maybe this could help: https://stackoverflow.com/questions/50363449/how-to-get-error-information-relating-to-typescript-file-lines-number-in-node-js.

tsconfig.json
Rohanator9000: disabled it in the newest tsconfig

tsconfig.json
```diff
@@ -0,0 +1,71 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
```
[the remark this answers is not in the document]
Rohanator9000: update the target to es2020 in the newest tsconfig

rvighne (approved):
@Rohanator9000 I ended up going back to ts-node; I realized it's not actually a whole separate compiler since it's supposed to be used exclusively in place of tsc when developing, so I updated the scripts to reflect that. It also handles dependencies better than nodemon.

PR #32, opened 2021-01-22
Job Sorting

erlstanleylee:
Topo Sort Added Heuristics:
When there are ties on which can be done first, do the one with the most outgoing edges.


alexzma (dismissed):
The build is failing, because the tester wants const instead of let for variables that don't get reassigned. Please change so it passes.
Could you also put toposort.ts into the src directory and toposort.test.ts into the src or test directory for organization purposes? Not sure about the directory for toposort.test.ts, because not sure if toposort.test.ts works with jest.

erlstanleylee:
Sorry about that forgot to actually move the files after uploading. 

Moved the .ts into src and moved .test.ts into the test file.

Rohanator9000:
@rvighne Could you provide some feedback on how I should handle the errors I'm throwing in `src/JobOrderer.ts`?

src/Client.ts
```diff
@@ -52,29 +51,81 @@ export default class Client extends EventEmitter {
...
+	 */
+	private checkJobsAndAssign(): void {
+		const job = this.jobOrderer.popNextJob()
+
+		if (job && this.availableDaemons.size > 0) {
+			const daemon = Array.from(this.availableDaemons.values())[0]
+			if (daemon) {
+				this.assignJobToDaemon(job, daemon)
+			} else {
+				throw 'skjdfhsadkfl'
```
rvighne, line 82: This condition is impossible, so don't check for it. It makes the code harder to read.

Also, consider allowing non-null assertions in the linter. That's a good concise way to indicate that something is always defined.

src/Client.ts
```diff
@@ -52,29 +51,81 @@ export default class Client extends EventEmitter {
...
+			} else {
+				this.jobOrderer.reportCompletedJob(job)
+			}
 
-		req.on('end', () => {
-			this.emit('progress', job, data)
-			this.available(client)
+			this.emit('progress', job.toString(), data)
+			this.setAvailable(daemon)
+			this.checkJobsAndAssign()
```
rvighne, line 128: This gets called twice (once within setAvailable). It's probably best to make setAvailable not call this, or change its name to make clear that it does.

src/Daemon.ts
```diff
@@ -11,6 +12,15 @@ function mockDelayMs(): number {
 	return MAX_DELAY_MS * Math.random()
 }
 
+/**
+ * Has a {@link FAILURE_PROBABILITY}% chance of returning true.
+ *
+ * @returns true {@link FAILURE_PROBABILITY}% of the time
+ */
+function percentChance(): boolean {
```
rvighne, line 20: This name is pretty unclear. Better: `mockFailure`

rvighne (changes requested):
General guidelines for errors:
- For runtime errors (something that can happen unpredictably, like network failure), return the error information to the caller explicitly. For example, reject the promise, have a special return value, or throw from an async function.
- For logic errors, you can throw an exception.

Things to fix:
- Always throw an instance of `Error` rather than strings.
- Don't check for errors that can never happen.
- Don't check for precondition violations; or if you do, make them assertions rather than throws.

src/JobOrderer.ts
```diff
@@ -0,0 +1,138 @@
...
+			job.incompletePrerequisites.forEach((prerequisite) => {
+				if (prerequisite == job) {
+					throw `Circular dependency: job ${job} has itself as a preqrequisite.`
+				} else {
+					const prerequisiteDependents = this.jobToDependents.get(prerequisite)
+
+					if (prerequisiteDependents) {
+						prerequisiteDependents.add(job)
+					} else {
+						throw `Job ${job} has prerequisite ${prerequisite} which was not passed to the constructor.`
```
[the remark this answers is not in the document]
rvighne: In fact, why not just take the root jobs (or just one root) as the argument? The job will be linked to everything it needs; it's already a tree. This avoids having redundant information.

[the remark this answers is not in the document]
Rohanator9000: If I were to use just the root job(s), each `Job` would need to know what all of its dependents are. Right now they know what all of their prerequisites are. I feel like the latter is the more intuitive of the two. We could include both, but that's redundant information.
Thoughts?

src/JobOrderer.ts
```diff
@@ -0,0 +1,138 @@
...
+	 *
+	 * @throws When one of the jobs has a prerequisite that is not one of the jobs, or is itself.
+	 * @param jobs - The jobs to manage. They must have correctly populated prerequisites fields.
+	 */
+	constructor(jobs: Job[]) {
+		// Initialize dependents map.
+		jobs.forEach((job) => this.jobToDependents.set(job, new Set()))
+
+		// Populate dependents map.
+		jobs.forEach((job) => {
```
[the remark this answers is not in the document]
Rohanator9000: Agreed. I'm adding https://github.com/github/eslint-plugin-github for https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md.

src/Job.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * A job for a Daemon to complete.
+ */
+export class Job {
+	public incompletePrerequisites: Set<Job>
```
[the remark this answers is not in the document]
Rohanator9000: By incomplete I mean unfinished. Incomplete prerequisites refers to all prerequisite jobs that are not yet complete (because `Job` forgets about each prerequisite as soon as its complete).
Would `unfinishedPrerequisites` or `incompletePrerequisiteJobs` be better?

src/Client.ts
Rohanator9000: lmao this might be the worst error message I've written

src/JobOrderer.ts
```diff
@@ -0,0 +1,138 @@
...
+
+	/**
+	 * Returns the next job that the JobOrderer would choose to run.
+	 * If there is no job that can be run, returns null.
+	 *
+	 * @returns The next job that would be run.
+	 */
+	public peekNextJob(): Job | null {
+		if (this.sources.size) {
+			return Array.from(this.sources.values()).reduce((job1, job2) => {
```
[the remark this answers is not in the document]
Rohanator9000: Yeah I thought about that and then forgot. I added https://www.npmjs.com/package/heap-js for this.

src/Client.ts
```diff
@@ -35,14 +15,33 @@ export type Job = string
...
+	 * String representation of the host and port together.
+	 *
+	 * @remarks
+	 * Needed because IPv6 addresses must be bracketed to disambiguate colons.
+	 *
+	 * @param host - hostname or IP address
+	 * @param port - port number
+	 * @returns connection string suitable for a URL
+	 */
+	private static hostAndPort(host: string, port: number): string {
```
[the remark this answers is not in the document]
Rohanator9000: Actually yeah I don't think that should be in the class, but shouldn't it go in a separate util file? I feel like it's weird just keeping it in `Client.ts` since its such a general function.

src/JobOrderer.ts
```diff
@@ -0,0 +1,138 @@
...
+	 */
+	constructor(jobs: Job[]) {
+		// Initialize dependents map.
+		jobs.forEach((job) => this.jobToDependents.set(job, new Set()))
+
+		// Populate dependents map.
+		jobs.forEach((job) => {
+			job.incompletePrerequisites.forEach((prerequisite) => {
+				if (prerequisite == job) {
+					throw `Circular dependency: job ${job} has itself as a preqrequisite.`
```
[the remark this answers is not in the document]
Rohanator9000: Which of those two would you recommend?

src/Job.ts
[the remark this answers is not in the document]
rvighne: > `Job` forgets about each prerequisite as soon as its complete

Hmm this means that Job is a mutable object, which can lead to code that's hard to reason about. It would be better if a given Job is frozen once created, and whoever is using it (like the Orderer) is responsible for tracking what's complete.

src/JobOrderer.ts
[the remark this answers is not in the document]
rvighne: For now I'm leaning towards the former, since that's easy and everything still works. If needed, we can go with the industrial-strength option later :)

src/Client.ts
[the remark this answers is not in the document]
rvighne: A separate `Utils.ts` library would be great if it is needed by multiple modules. Until then, there's no harm in it being in `Client.ts` as a free function. It's normal for code to have private implementation details.

src/JobOrderer.ts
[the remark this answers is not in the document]
rvighne: > each `Job` would need to know what all of its dependents are.

Why? Isn't calculating dependents the responsibility of whoever is using a Job in that way (i.e. the JobOrderer)?

> Right now they know what all of their prerequisites are. I feel like the latter is the more intuitive of the two.

I agree, I believe this is possible without breaking that aspect, let me know if I've overlooked something.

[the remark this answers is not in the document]
Rohanator9000: By 'root job' I thought you meant the first `Job` to be completed, but now I see that you mean the final job.

Yeah that's a good point, I'll change the constructor to take just the root job.

src/Job.ts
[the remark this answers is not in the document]
Rohanator9000: Why do you think that a `Job` being mutable would make the code harder to reason about?

Rohanator9000:
@erlstanleylee Can you review this? I can't request it because you opened the PR.

src/Job.ts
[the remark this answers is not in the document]
rvighne: In general, when you have an object that can change, you always have to think about all the states it can be in. On the other hand, an immutable object needs to be constructed correctly, and then it's always correct. That's why a lot of functional programming languages require immutable data, and in JS we use `const` instead of `let` unless needed.

Of course we're not really trying to write a pure functional program here, so mutation is ok 😂. The distinction is between data, and machinery that acts on data. `Job` is data, so it should be unchangeable. Meanwhile JobOrderer has a bunch of private fields that are like the temporary storage of an algorithm, so it's totally fine to use that as scratch space.

Practically, you can fix this by having JobOrderer copy the data it needs and then modify the copy, and keep the modified stuff private. I havent't read the code in detail since the changes though, so let me know if you'd like more specific advice.

alexzma (approved):
This looks good to me. I assume the mockFailure in Daemon.ts is to test whether the client can handle failure, and won't be in the final product.

src/Http2Client.ts
```diff
@@ -40,38 +44,82 @@ export class Http2Client extends EventEmitter implements Client {
...
-	private available(client: ClientHttp2Session): void {
-		const job = this.queue.pop()
-		if (!job) {
-			client.close(() => this.emit('done'))
-			return
+	private checkJobsAndAssign(): void {
+		const job = this.jobOrderer.popNextJob()
+
+		if (job && this.availableDaemons.size > 0) {
+			const daemon: ClientHttp2Session = this.availableDaemons.values().next()
```
rvighne, line 73: Take a randomly chosen daemon, not the next available. That's an easy way to ensure we distribute jobs evenly among the daemons.

src/controller.ts
```diff
@@ -16,13 +19,15 @@ function done() {
...
-	'first',
-])
+const job3: NormalJob = new NormalJob('third')
+const job4: NormalJob = new NormalJob('fourth')
+const job2: NormalJob = new NormalJob('second', new Set([job3]))
+const job1: NormalJob = new NormalJob('first', new Set([job4, job2]))
+const job5: NormalJob = new NormalJob('fifth', new Set([job1]))
+
+const client: Client = new Http2Client(
+	new HeapJobOrderer([job1, job2, job3, job4, job5]),
```
rvighne, line 29: Some of these are not root jobs. Is that intentional?

src/Http2Client.ts
```diff
@@ -40,38 +44,82 @@ export class Http2Client extends EventEmitter implements Client {
...
+	/**
+	 * Asks the daemon to work on the job.
+	 * Sends HTTP request and handles response.
+	 *
+	 * @param job - The job to assign.
+	 * @param daemon - The daemon to which to assign the job.
+	 */
+	private assignJobToDaemon(job: Job, daemon: ClientHttp2Session) {
+		const request = daemon.request({ ':path': `/${job.getName()}` })
+		daemon.on('error', (err) => this.emit('error', err))
```
rvighne, line 108: Instead of emitting error, we can emit a less disastrous event (or none at all) and handle this within the Client. The error handler could
1. Tell JobOrderer to "reportInconclusiveJob" (I just made that up) which puts it back as if we had never tried it. Or have some other way of holding on to this job.
2. Implement some policy re. the daemon; in the simplest case we could just close the connection and remove the daemon from the list.
3. Call checkJobsAndAssign

Thoughts?

rvighne (changes requested):
This is great, I like that you added to the HttpClient because we can reuse a lot of this code for the SSH client later. I think the non-networking part of the client is basically done after your change 🎉

The significant comments are about the unit tests and some minor bugs in Client. Also, I haven't gotten the chance to review HeapJobOrderer yet.

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+	private jobToCompletedPrereqs: Map<Job, Set<Job>> = new Map<Job, Set<Job>>()
+
+	/**
+	 * Create the JobOrderer and inform it of the jobs to manage.
+	 *
+	 * @throws When one of the jobs has a prerequisite that is not one of the jobs, or is itself.
+	 * @param jobs - The jobs to manage. They must have correctly populated prerequisites fields.
+	 */
+	constructor(rootJobs: Job[]) {
+		// Initialize the sources heap with its sorting function.
```
rvighne, line 27: In this comment, describe what the sorting function does (I'm guessing descending by number of dependents?) and why.

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+		for (const rootJob of rootJobs) {
+			this.jobToDependents.set(rootJob, new Set())
+		}
+
+		const queue: Job[] = rootJobs
+		const seenSources: Set<Job> = new Set() // We use this because checking whether a heap has an element is O(n).
+
+		// BFS traverse dependency graph, starting with root nodes (the ultimate jobs).
+		while (queue.length) {
+			const job = queue.pop()
```
rvighne, line 50: This will modify the `rootJobs` array passed in as an argument; you probably want to make a copy first.

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+			assert(job, 'Queue with nonzero length should have returned object.')
+
+			// Don't process the same node twice.
+			if (!seenSources.has(job) && !this.nonSources.has(job)) {
```
rvighne, line 54: Can you explain why there's two disjoint sets (seenSources, nonSources) and a heap (sourcesHeap) here? I.e. why not just have one set for holding all the nodes, not just sources (nice side effect: it would catch duplicates of any kind, so you get cycle detection for free) and another priority queue to store the sources?

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+export class HeapJobOrderer implements JobOrderer {
+	private sourcesHeap: Heap<Job>
+
+	private nonSources: Set<Job> = new Set()
+
+	private inProgress: Set<Job> = new Set()
+
+	private jobToDependents: Map<Job, Set<Job>> = new Map<Job, Set<Job>>()
+
+	private jobToCompletedPrereqs: Map<Job, Set<Job>> = new Map<Job, Set<Job>>()
```
rvighne, line 18: Documentation for all these fields would help others understand the algorithm.

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+	 * JobOrderer will update the dependency graph and then forget about this job.
+	 *
+	 * @param completedJob - The job that is completed.
+	 */
+	public reportCompletedJob(completedJob: Job): void {
+		const dependents = this.jobToDependents.get(completedJob)
+
+		if (!dependents) {
+			throw new Error(
+				`We don't know about this job marked completed: ${completedJob.toString()}.`,
```
rvighne, line 108: Good use of error handling here.

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+		for (const dependent of dependents) {
+			const dependentCompletedJobs = this.jobToCompletedPrereqs.get(dependent)
+
+			assert(
+				dependentCompletedJobs,
+				`Job ${completedJob.toString()} has dependent that we didn't process, which should be impossible.`,
+			)
+			dependentCompletedJobs.add(completedJob)
+
+			if (dependent.getNumPrerequisites() === dependentCompletedJobs.size) {
```
rvighne, line 121: Add an assertion that completedJob is a member of dependent's prerequisites. Otherwise it could take a while to convince yourself that this line is equivalent to "dependent's prereqs and dependentCompletedJobs are the same set"

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+				`We don't know about this job marked completed: ${completedJob.toString()}.`,
+			)
+		}
+
+		for (const dependent of dependents) {
+			const dependentCompletedJobs = this.jobToCompletedPrereqs.get(dependent)
+
+			assert(
+				dependentCompletedJobs,
+				`Job ${completedJob.toString()} has dependent that we didn't process, which should be impossible.`,
```
rvighne, line 117: Don't call toString manually, as it will get automatically called when converting to string (and there are other ways than toString which JS can use to do this conversion).

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+				this.nonSources.delete(dependent)
+			}
+		}
+
+		this.inProgress.delete(completedJob)
+	}
+
+	/**
+	 * Marks an in-progress job as failed.
+	 * JobOrderer will try to run it again at some point.
```
rvighne, line 132: Ah interesting. We have two kinds of failures:
- network failure, in which case we should retry (so call reportFailedJob)
- the job actually exited with non-0 status, in which case we should call off the whole thing. I had originally thought that's what reportFailedJob does, but actually we can just abandon the JobOrderer in that case.

rvighne (commented):
Finishing up HeapJobOrderer which I didn't get to yesterday.

src/Http2Client.ts
rvighne: After reading HeapJobOrderer, I realize this is already in place (all you need to do is call reportFailedJob from this error handler). No need to implement it in this PR.

src/HeapJobOrderer.ts
rvighne: Probably a good idea to test this error in the unit tests.

src/Daemon.ts
```diff
@@ -21,18 +31,25 @@ export function createDaemon(): Server {
...
 
 		const output = input.toUpperCase()
 		setTimeout(() => {
-			stream.respond({
-				':status': 200,
-				'content-type': 'text/plain',
-			})
-			stream.end(output)
+			if (mockFailure()) {
+				stream.respond({ status: 400, 'content-type': 'text/plain' })
```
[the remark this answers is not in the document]
Rohanator9000: nitpicky is good!

src/NormalJob.ts
```diff
@@ -0,0 +1,49 @@
...
+	public getNumPrerequisites(): number {
+		return this.prerequisites.size
+	}
+
+	/**
+	 * JavaScript calls this function whenever Job is casted to a string.
+	 *
+	 * @returns This job's name.
+	 */
+	public toString(): string {
```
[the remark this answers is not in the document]
Rohanator9000: I call `getName()` so `toString()` isn't called on the prerequisites.

Also, we can just print the name, or number of prerequisites, but the default `toString()` returns `[object Object]`, so I don't think we should use that.

src/NormalJob.ts
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * @param name - The job's name. Must be unique between jobs in the same dependency graph.
+	 * @param prerequisites - An optional array of jobs that must be completed before this job can start.
+	 */
+	constructor(
+		private name: string,
+		private prerequisites: Set<Job> = new Set(),
+	) {}
+
+	public getName(): string {
```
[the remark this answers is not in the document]
Rohanator9000: I usually like to be explicit about visibility because it 1) forces me to think about it and 2) shows my intent better (I think). But it's nbd so I can remove them if you feel strongly about it.

src/JobOrderer.ts
```diff
@@ -0,0 +1,8 @@
+import { Job } from './Job'
+
+export interface JobOrderer {
+	popNextJob(): Job | null
```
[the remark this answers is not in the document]
Rohanator9000: That doesn't work, and I couldn't find another way. I know you can do `property?: value` but that's if the property can be `null`.

.eslintrc.json
```diff
@@ -17,10 +19,20 @@
...
 		"import/no-default-export": "error",
-		"tsdoc/syntax": "error"
+		"tsdoc/syntax": "error",
+		"no-restricted-syntax": [
+			"error",
+			"ForInStatement",
+			"LabeledStatement",
+			"WithStatement"
+		],
+		"import/order": "off",
```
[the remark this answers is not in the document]
Rohanator9000: This was conflicting with another rule about order preference in regards to `import assert = require('assert')`. But I've removed that so I'll add this rule back.

src/Http2Client.ts
```diff
@@ -40,38 +44,82 @@ export class Http2Client extends EventEmitter implements Client {
...
 	 */
-	private available(client: ClientHttp2Session): void {
-		const job = this.queue.pop()
-		if (!job) {
-			client.close(() => this.emit('done'))
-			return
+	private checkJobsAndAssign(): void {
+		const job = this.jobOrderer.popNextJob()
+
+		if (job && this.availableDaemons.size > 0) {
```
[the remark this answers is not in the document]
Rohanator9000: No because if no Jobs are currently available, none will become available until another Job completes. If a job completes then the Daemon must have responded to our request. Every time we get a response we run `setAvailableAndCheckJobs()`.

src/Http2Client.ts
```diff
@@ -40,38 +44,82 @@ export class Http2Client extends EventEmitter implements Client {
...
+	}
+
+	/**
+	 * Asks the daemon to work on the job.
+	 * Sends HTTP request and handles response.
+	 *
+	 * @param job - The job to assign.
+	 * @param daemon - The daemon to which to assign the job.
+	 */
+	private assignJobToDaemon(job: Job, daemon: ClientHttp2Session) {
```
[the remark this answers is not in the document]
Rohanator9000: Yeah you're right I'll add that.

And why not just use a map from daemon to num of concurrent jobs?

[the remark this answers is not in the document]
Rohanator9000: Oh I just saw the random thing.

src/Http2Client.ts
Rohanator9000: Getting a random element from a `Set` is O(n). I could have an array or something that also contains the daemons for random access.

What do you mean by evenly? Do you think that the daemons will come to us in a significantly nonrandom order?

src/HeapJobOrderer.ts
erlstanleylee: Initially we were going to do what you were suggesting, but we left it as two disjoint sets to allow an further optimizations that we may want to do with deciding which job. This is an optional feature that we may or may not implement.

src/HeapJobOrderer.ts
```diff
@@ -0,0 +1,156 @@
...
+				for (const prerequisite of job.getPrerequisitesIterable()) {
+					let dependents = this.jobToDependents.get(prerequisite)
+
+					if (!dependents) {
+						dependents = new Set()
+						this.jobToDependents.set(prerequisite, dependents)
+					}
+
+					dependents.add(job)
+					queue.push(prerequisite)
```
erlstanleylee, line 65: I might be missing something but why are we pushing the prerequisite back onto the queue. Wouldn't the queue already have all the jobs anyway? There is also the check for whether we have processed the same node above.

erlstanleylee:
@Rohanator9000 I reviewed the JobOrderer stuff and what I could to the rest. Just 1 minor comment. Looks good!

src/Http2Client.ts
```diff
@@ -40,38 +44,82 @@ export class Http2Client extends EventEmitter implements Client {
...
+			if (data === 'failed') {
+				this.jobOrderer.reportFailedJob(job)
+			} else {
+				this.jobOrderer.reportCompletedJob(job)
+			}
 
-		req.on('end', () => {
-			this.emit('progress', job, data)
-			this.available(client)
+			this.emit('progress', job.toString(), data)
```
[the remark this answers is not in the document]
Rohanator9000: I didn't have to.

src/Http2Client.ts
rvighne: > Getting a random element from a `Set` is O(n).

Good point, I'm not sure atm how to cleanly solve this. (since this isn't performance critical, it's ok to do the O(n) operation for now with a comment stating the future optimization).

> Do you think that the daemons will come to us in a significantly nonrandom order?

The daemons may come to us in any order, and that order probably isn't random. Besides, even if it is, the issue is that the order is stable (unless the Set ordering is internally randomized like with Python's dict, in which case this is less important).

So as a long-running DAG is being worked on, machine 1 would get 100% utilization, machine 2 only when machine 1 is busy so maybe 50% depending on the amount of parallelism in the graph, and so on with the last machine getting no use (for example).

rvighne: Ah I just found the [MDN docs for Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set):
> You can iterate through the elements of a set in insertion order.

That's interesting. I guess that means that we would pop the next available daemon from the front, then return it to the back when we're done. You might want to confirm this is really what happens when you remove() and re-add(). If so, perhaps that gives us enough of an even-distribution guarantee?

.eslintrc.json
```diff
@@ -17,10 +19,20 @@
 		"prettier/@typescript-eslint"
 	],
 	"rules": {
-		"no-console": "off",
+		"github/no-then": "off", // TODO enable.
+		"import/no-namespace": "off", // TODO enable.
+		"no-console": "off", // TODO enable.
```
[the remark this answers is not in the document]
Rohanator9000: I actually don't need to disable `import/no-namespace`, I think I removed the error-causing code. For `github/no-then` we need to stop using `.then()`. For `no-console` we need to stop logging using `console`.

.eslintrc.json
```diff
@@ -8,6 +8,8 @@
 	"extends": [
 		"eslint:recommended",
 		"plugin:@typescript-eslint/recommended",
+		"plugin:github/recommended",
+		"plugin:github/typescript",
```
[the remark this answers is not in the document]
Rohanator9000: Oh yeah I only needed one rule from GitHub so I just included that directly and removed the config extension. But I still need to override some rules.

src/HeapJobOrderer.ts
Rohanator9000: We've seen `job` but not necessarily `job`'s prerequisites.

src/HeapJobOrderer.ts
Rohanator9000: The only way that job A can become a dependent of job B is if job B is in job A's prerequisites. I can add the `assert`, but isn't that kind of trivially true?

src/HeapJobOrderer.ts
Rohanator9000: That's what I thought, but https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/restrict-template-expressions.md. I'm not a huge fan of this rule so I wouldn't mind turning it off.

src/HeapJobOrderer.ts
Rohanator9000: `nonSources` is a field that I use at other points in the code. I use `seenSources` because I need to a way to check whether I've already seen a source and I don't wanna check the heap because that's O(n).

Also I don't see how turning `nonSources` to `allJobs` would catch cycles.

test/HeapJobOrderer.test.ts
```diff
@@ -0,0 +1,51 @@
+import { NormalJob } from '../src/NormalJob'
+import { HeapJobOrderer } from '../src/HeapJobOrderer'
+
+/**
+ * Testcase with 5 jobs.
+ */
+test('HeapJobOrderer', () => {
```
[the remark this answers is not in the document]
Rohanator9000: I'm adding other test cases for more granular information. For this test case, my intention is for it to test whether the JobOrderer correctly orders Jobs. I can remove the `expect` calls that use `isDone()`, but all the other `expects` seem related to it correctly ordering Jobs.

Thoughts?

src/Http2Client.ts
Rohanator9000: I don't think that we should depend on `Set`'s implementation as that can change at any time.

Also, if you want to equally distribute work, it could be better to actually keep track of how many jobs each Daemon has and assign using that. Delegating to randomness seems kinda inefficient.

Regardless, I think that this should be part of another PR, since the focus of this PR isn't even the Client.

src/NormalJob.ts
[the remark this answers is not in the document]
rvighne: Ah makes sense. I was thinking of console.log which shows the indented view instead of [object Object]

src/Http2Client.ts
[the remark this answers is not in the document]
rvighne: I'm thinking of the case where one job finishes and unlocks multiple other jobs that were depending on it. In that case, the job finishing will call this function, but it will only pick up one of the potentially many jobs now available.

src/Http2Client.ts
rvighne: > I don't think that we should depend on `Set`'s implementation as that can change at any time.

It's not an implementation detail of Set, it appears to be part of its specified behavior, so it's fine to rely on it. No changes are necessary though, the current code most likely does what we want (pops from the "front" of the Set in insertion order, pushes to the "back").

> Delegating to randomness seems kinda inefficient.

It would actually be more efficient - you end up distributing evenly without having to maintain any data structures. But as explained above it's not needed here because of the way Set works.

src/HeapJobOrderer.ts
rvighne: Interesting. Let's leave it as is then.

src/HeapJobOrderer.ts
rvighne: I see. It would help readability to rename things like so:
- sourcesHeap.
- seenSources -> sourcesSet (to show that it's the same data as sourcesHeap, just as a set; a comment stating this would also be good).
- nonSources -> nonSourcesSet (I don't usually like putting the type in the name, but this way you can tell it's like a parallel to sourcesSet).

Any better suggestions for the names? I'm not totally happy with my suggestions here but I do think they need to be made more clear.

> Also I don't see how turning nonSources to allJobs would catch cycles.

Any time you traverse a tree and add all the elements to a set, if you ever try to add an element that's already there, you know that's a cycle.

src/Http2Client.ts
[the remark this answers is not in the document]
Rohanator9000: Yeah you're totally right. I just pushed some code to fix it.

.eslintrc.json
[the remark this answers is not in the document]
rvighne: A rule similar to `github/no-then` is already included in `eslint-plugin-promises` so we can probably remove it here. Also it doesn't always make sense to avoid "then", especially in top-level code where you can't use `await`.

For no-console: good point. When you make the PR to add directory structure, we can go ahead and remove this override, cause it's only needed for our program (non-module) code.

src/UnkownJobError.ts
```diff
@@ -0,0 +1 @@
+export class UnknownJobError extends Error {}
```
rvighne, line 1: Spelling (of this file name)

test/NormalJob.test.ts
```diff
@@ -0,0 +1,45 @@
+import { Job } from '../src/Job'
+import { NormalJob } from '../src/NormalJob'
+
+describe('NormalJob', () => {
+	it('returns the correct name', () => {
+		const job = new NormalJob('jobName')
+		expect(job.getName()).toEqual('jobName')
```
rvighne, line 7: Make this name a constant since you use it in two places.

test/NormalJob.test.ts
```diff
@@ -0,0 +1,45 @@
+import { Job } from '../src/Job'
+import { NormalJob } from '../src/NormalJob'
+
+describe('NormalJob', () => {
```
rvighne, line 4: Good test cases.

test/NormalJob.test.ts
```diff
@@ -0,0 +1,45 @@
...
+	it('returns the correct name', () => {
+		const job = new NormalJob('jobName')
+		expect(job.getName()).toEqual('jobName')
+	})
+
+	it('returns the correct number of prerequisites', () => {
```
rvighne, line 10: `sourceJob`, `nonSourceJob` should be separate test cases.

src/HeapJobOrderer.ts
Rohanator9000: Hmm I also don't like putting the type in the name, which makes me want to change `sourcesHeap`. How's this:

- `sourcesHeap` -> `sources`
- `seenSources` -> `sourcesHelperSet` (with a comment)
- `nonSources`

This way we still have the parallelism between sources and nonSources. This also indicates that `sourcesHelperSet` is a temporary helper variable.

test/NormalJob.test.ts
```diff
@@ -0,0 +1,45 @@
...
+			new NormalJob('3'),
+		])
+
+		const job = new NormalJob('job', prerequisites)
+
+		for (const prerequisite of job.getPrerequisitesIterable()) {
+			prerequisites.delete(prerequisite)
+		}
+
+		expect(prerequisites.size).toEqual(0)
```
rvighne, line 43: Simpler test: just create a set from this iterable, then expect it to be equal to `prerequisites`.

That would also work in case NormalJob modifies prerequisites (I know it doesn't, because of the previous test case, but better to be safe).

src/HeapJobOrderer.ts
Rohanator9000: I think my original thinking was that using a set would prevent any errors if someone called `reportCompletedJob()` twice with the same Job. But I can just move the `inProgress()` existence-check up to prevent this.

So yeah I'll just store the number. Great catch.

rvighne (changes requested):
Looks good overall. Code coverage report:
```
-------------------|---------|----------|---------|---------|-------------------
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------|---------|----------|---------|---------|-------------------
All files          |   92.05 |    89.29 |   94.44 |   91.76 |
 Constants.ts      |     100 |      100 |     100 |     100 |
 HeapJobOrderer.ts |   87.93 |       88 |   85.71 |   87.93 | 116-124,189-194
 NormalJob.ts      |     100 |      100 |     100 |     100 |
 UnkownJobError.ts |     100 |      100 |     100 |     100 |
 ZeroconfClient.ts |     100 |      100 |     100 |     100 |
 ZeroconfDaemon.ts |     100 |      100 |     100 |     100 |
-------------------|---------|----------|---------|---------|-------------------
```
Might want to look into those lines and test them if possible.

src/HeapJobOrderer.ts
rvighne: Yeah this is even better.

test/HeapJobOrderer.test.ts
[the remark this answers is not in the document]
rvighne: Fair enough, I agree popNextJob is hard to test in isolation. With the other test cases you added it looks good now.

rvighne:
Also btw, great documentation for the Job and JobOrderer interfaces.

Rohanator9000:
> Looks good overall. Code coverage report:
> 
> ```
> -------------------|---------|----------|---------|---------|-------------------
> File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
> -------------------|---------|----------|---------|---------|-------------------
> All files          |   92.05 |    89.29 |   94.44 |   91.76 |
>  Constants.ts      |     100 |      100 |     100 |     100 |
>  HeapJobOrderer.ts |   87.93 |       88 |   85.71 |   87.93 | 116-124,189-194
>  NormalJob.ts      |     100 |      100 |     100 |     100 |
>  UnkownJobError.ts |     100 |      100 |     100 |     100 |
>  ZeroconfClient.ts |     100 |      100 |     100 |     100 |
>  ZeroconfDaemon.ts |     100 |      100 |     100 |     100 |
> -------------------|---------|----------|---------|---------|-------------------
> ```
> 
> Might want to look into those lines and test them if possible.

Cool, what did you use to make that report?

test/NormalJob.test.ts
Rohanator9000: I don't see a way to check `Set` equality that's more efficient than this. Since this is just a test, it could still be better if its more elegant/readable, but unfortunately I haven't found an elegant way to check `Set` equality.

.eslintrc.json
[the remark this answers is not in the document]
Rohanator9000: Since I stopped extending GitHub, I don't even need to override it anymore.

rvighne:
@Rohanator9000 
```
npx jest --coverage
```

test/NormalJob.test.ts
rvighne: I believe `toEqual` in jest does deep equality, so it can handle any JS builtin types (including Set).

Rohanator9000:
> Looks good overall. Code coverage report:
> 
> ```
> -------------------|---------|----------|---------|---------|-------------------
> File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
> -------------------|---------|----------|---------|---------|-------------------
> All files          |   92.05 |    89.29 |   94.44 |   91.76 |
>  Constants.ts      |     100 |      100 |     100 |     100 |
>  HeapJobOrderer.ts |   87.93 |       88 |   85.71 |   87.93 | 116-124,189-194
>  NormalJob.ts      |     100 |      100 |     100 |     100 |
>  UnkownJobError.ts |     100 |      100 |     100 |     100 |
>  ZeroconfClient.ts |     100 |      100 |     100 |     100 |
>  ZeroconfDaemon.ts |     100 |      100 |     100 |     100 |
> -------------------|---------|----------|---------|---------|-------------------
> ```
> 
> Might want to look into those lines and test them if possible.

@rvighne Fixed this, 100% coverage.

rvighne (approved):
🚀🚀

PR #34, opened 2021-01-24
Implement zero-config network discovery and mock client/daemon pair

rvighne:
This uses the Zeroconf protocol to find peers on the LAN. The version of Zeroconf being used here is Bonjour. I have tested it locally using Windows.

To put this in a practical context, this PR also adds a dummy client-server pair. These use HTTP/2 for simplicity. The server is very dumb and will probably be totally replaced. On other hand, the client comes with a decent implementation of sending jobs from a queue to multiple servers while more servers dynamically join, so it can be the basis of a more complete implementation.

## How to test
In any order:
1. Run as many servers as you want (`npm start`), either on the same machine or different machines on the same LAN.
2. Run a client (`npm run cli`). It should find all the servers you started (log their service info to the console). Then, it should finish all 5 of its tasks, while server consoles will show that different servers got the requests.

Sample output

## Interesting features
- The Zeroconf code is deliberately kept separate from the client/server.
- The daemon doesn't specify a port to listen on, so the OS assigns it a port. This makes it possible to run multiple daemons on one machine and do local testing. And it makes it so you *have* to use Zeroconf to know where the server is 😀

## Misc
This PR also adds a documentation generator (because I added doc-strings to the methods). You can view it in `docs/index.html`.

## Next steps
- Test using daemons on multiple physical machines.
- Write integration tests.


alexzma (approved):
It works. Just in case others need to know, the example is run by dist/controller.js and dist/server.js after compilation.

rvighne:
@alexzma Thanks, I added some testing instructions now. Btw, which OS did you test it on?

alexzma:
> @alexzma Thanks, I added some testing instructions now. Btw, which OS did you test it on?

I tested it on MacOS.

src/controller.ts
```diff
@@ -1,2 +1,24 @@
-console.error('Not implemented')
-process.exit(1)
+#!/usr/bin/env node
```
Rohanator9000, line 1: Why do we need this shebang?

src/Client.ts
```diff
@@ -1,5 +1,80 @@
...
+	}
+
+	private available(client: ClientHttp2Session): void {
+		const job = this.queue.pop()
+		if (!job) {
+			client.close(() => this.emit('done'))
+			return
+		}
+
+		const req = client.request({ ':path': `/${job}` })
```
Rohanator9000, line 68: Why did you choose to call this field `:path`? Will jobs be represented as paths?

src/controller.ts
```diff
@@ -1,2 +1,24 @@
...
+const zeroconf = bonjour()
+
+function done() {
+	console.log('Finished')
+
+	// The mDNS socket apparently has no way to tell that it's not needed.
+	zeroconf.destroy()
+}
+
+const client = new Client(['fifth', 'fourth', 'third', 'second', 'first'])
```
Rohanator9000, line 18: How would this work if someone wants multiple clients? Would there be multiple instances of this controller running?

test/Client.test.ts
```diff
@@ -1,5 +0,0 @@
-/* TODO: real unit tests */
```
Rohanator9000, line 1: Why delete this file (and `Daemon.test.ts` too)?

src/controller.ts
rvighne: This makes the script directly executable as if it's a binary. This makes it work with npx (so you can do `npx junknetd` or `npx junknet-make`). Also, it's good as a signal that this file is an entrypoint to our program as opposed to library code.

src/Client.ts
rvighne: I didn't choose that name, `:path` is the HTTP/2 equivalent of the path component of a URL.

We won't be representing jobs as paths in the future (although... could be an interesting idea). This is just because I'm using HTTP as a stand-in for our more complex protocol, which will be [something like this](https://github.com/cs130-w21/1/wiki/Network-Protocol).

src/controller.ts
rvighne: Sure, how you use the `Client` is up to you ("you" being the user of the client library). I think the controller would probably just use one client and exit when it completes, but it's flexible. (instances of the client class are independent from each other).

test/Client.test.ts
rvighne: Good question 😂 I think I didn't intend to delete it but I renamed them a few times and and they got lost; I'll add it back. (Though not with actual testcases yet, I want to add tests only once we switch away from this HTTP demo to our actual protocol).

tsconfig.json
```diff
@@ -4,7 +4,7 @@
 		"target": "ES2020",
 		"module": "commonjs",
 		"allowJs": false,
-		"sourceMap": true,
```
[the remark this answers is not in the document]
rvighne: It's not needed, since during development we should be using `ts-node` (via `npm start` or `npm run cli`), which always shows source locations. Though it can't hurt, so I'll add it back in.

test/Client.test.ts
rvighne: Actually, Jest doesn't let you have a suite without testcases, and the current dummy testcase isn't valid (since the code it "tested" is gone). Rather than create yet another dummy testcase I'd rather delete it; we definitely will be adding unit tests to all files in the future.

tsconfig.json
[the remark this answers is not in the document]
Rohanator9000: Good point. I don't mind it being included or not since it doesn't matter, so up to you.

Rohanator9000 (approved):
Great work, I appreciate your clarity in TS docs, it makes reading code much easier.

Issue #14, opened 2021-01-12
Daemon executes tasks


rvighne:
## User Story
As a developer, I want this daemon to receive work and send the results back so that I can use it to execute sub-tasks of a job.

## Detailed Description
As a user, I would like to send input into the daemon and receive output from the daemon so that I can process the inputs and have the results of the processing.

## Acceptance Criteria
- [ ] Given input file(s), transfer the contents of the input file(s) into the docker container.
- [ ] Given output file(s), output the appropriate results of the docker command to the output file(s).
- [ ] Given an input command and container, run the command within the container.

Estimated points: 4

PR #35, opened 2021-01-25
Dockerode exec

alexzma:
DockerExec.ts has an API that allows importing of images and creating, stopping, running, and removing containers.

listContainers lists the containers.
listImages lists the images.
importImage imports an image with a given name.
createContainer creates a container with given image, command, and volumes.
attachStreams attaches given input, output, and error streams to a given container.
stopContainer stops a given container.
removeContainer removes a given container.

There is also a function called runCommand in docker-demo.ts that is an example of how a command can be run on a container.


src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+				}
+				console.log('container end: ', data)
+				await container.remove()
+				console.log('removed')
+			})
+		}
+	})
+}
+
+//This method is only an example of how to run a command. This is not part of the API
```
rvighne, line 146: This is a great example, we should avoid leaving code in comments though. Can you move this into a separate file like `src/docker-demo.ts` and add a `#!` line so we can directly run it?

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+	try {
+		const stream = await docker.pull(name)
+		return stream
+	} catch (err) {
+		console.log(err)
+	}
+}
+
+//volumePairs should be of the form [[hostSrc,containerDest]]
+const createContainer = async (
```
rvighne, line 37: This function (actually all, but especially this one) needs documentation. See #34 for an example of TSDoc comments (these will be automatically shown by vscode to anyone who uses this function).

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+		console.log(err)
+		return []
+	}
+}
+
+const listContainers = async () => {
+	try {
+		const containers = await docker.listContainers()
+		return containers
+	} catch (err) {
```
rvighne, line 21: We really shouldn't be ignoring this error (same applies to the other functions like this). It's better to let the caller know so they can handle it properly.

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+		console.log(err)
```
rvighne, line 22: Since this is a library, it shouldn't print anything to console (same applies to the other functions like this).

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+	command: string[],
+	inputDir: string,
+	stdinFile: string,
+	outputDir: string,
+	stdoutFile: string,
+	errorDir: string,
+	stderrFile: string,
+	volumePairs: [string, string][],
+) => {
+	try {
```
rvighne, line 48: The gigantic try-catch is a bad idea, we're ignoring a lot of important and unexpected errors. This function should have a way to report errors, and/or be broken up into multiple internal functions.

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+		const stream = await container.attach({
+			hijack: true,
+			stream: true,
+			stdin: true,
+			stdout: true,
+			stderr: true,
+		})
+		let stdoutStream: any = process.stdout
+		if (outputDir != '' && stdoutFile != '') {
+			await fs.mkdir(outputDir, { recursive: true }, (err) => {
```
rvighne, line 77: Creating directories and writing files is distracting from the main point of this library. Why not just return the `Stream`s instead of taking dirs/files as arguments? That way the caller can do anything with them, such as forward them to the client directly through a socket. They can still log them to disk if they need that.

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+					console.log(err)
+					throw err
+				}
+			})
+			stderrStream = fs.createWriteStream(`${errorDir}/${stderrFile}`)
+		}
+		container.modem.demuxStream(stream, stdoutStream, stderrStream)
+
+		if (inputDir != '' && stdinFile != '') {
+			const readStream = fs.createReadStream(
```
rvighne, line 98: This is an interesting idea, but we will probably not need to give `stdin` to the recipe at all, since Make doesn't do that.

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
+import fs = require('fs')
+//import path = require('path')
+import Docker = require('dockerode')
+//needs docker to already be running
+const docker = new Docker()
+
+const listImages = async () => {
```
rvighne, line 7: Don't use arrow functions at the top level. They make it optional to specify the return type, but it's important for library functions to have a clear type signature so that they're self-documenting.

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		return container
+	} catch (err) {
+		console.log(err)
+		return null
+	}
+}
+
+const stopContainer = async (container: any) => {
```
rvighne, line 112: Don't use `any` (unless typescript is being dumb and it's impossible to compile without it). Use a Container interface or type if there is one.

rvighne (changes requested):
Very cool, this pretty much completes our integration with Docker 👌

src/DaemonExec.ts
```diff
@@ -0,0 +1,244 @@
+import fs = require('fs')
+//import path = require('path')
+import Docker = require('dockerode')
```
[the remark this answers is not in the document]
alexzma: I tried importing it that way, but the fs was failing. For some reason, fs only works with dynamic import.

src/DaemonExec.ts
alexzma: The stdin is for the container command. For if the makefile had a command like "process <input >output", where input and output would be in the listed volumes.

rvighne: That would actually be taken care of by the shell, since Make starts a subshell for each recipe line. Still, it's fine to leave this in just in case we think of a future use for it.

src/DaemonExec.ts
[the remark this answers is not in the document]
rvighne: Hmm that's odd, I'll look into it later.

pjot726 (approved):
Looks good to me, I approve from my end.

pjot726:
@alexzma  I don't think this PR will solve #14

alexzma:
> @alexzma I don't think this PR will solve #14

I think you're right. I'll change it.

rvighne:
@alexzma I enabled type-aware linting (I think this is how we intended to use ESLint, but it's disabled by default) and the most popular style guide for JS, which caught a lot of the issues I mentioned in comments. Apologies for the now-massive amount of linting errors that weren't there before 😁

I also got the static imports working and fixed bugs which the linter found in code that's already in master.

src/DaemonExec.ts
```diff
@@ -0,0 +1,138 @@
+import * as Docker from 'dockerode'
+
+interface ContainerModem {
+	demuxStream(
+		stream: NodeJS.ReadWriteStream,
+		stdoutStream: NodeJS.WritableStream,
+		stderrStream: NodeJS.WritableStream,
+	): void
+}
```
rvighne, line 9: Avoid repeating this interface definition both here and in the demo script; can you merge them and move it to a separate file (e.g. `DockerodeTypes.ts`)? Or, even better, make a PR to `@types/dockerode` 😉.

src/DaemonExec.ts
```diff
@@ -0,0 +1,138 @@
...
+ * @param image - the image the container uses
+ * @param command - an array that holds the command for the container
+ * @param volumePairs - an array of string pairs that represents the volumes
+ * @returns the Container
+ */
+export async function createContainer(
+	docker: Docker,
+	image: string,
+	command: string[],
+	volumePairs: [string, string][] = [],
```
rvighne, line 61: Looking at this function, the caller doesn't know what to provide here (I'm assuming it's 2-arrays of `[from path, to path]`). Can you export an interface like:

```ts
interface VolumeDefinition {
	fromPath: string
	toPath: string
}
```

And use an array of that as the parameter? This would also require changing the use of this function to use an object with named keys rather than a 2-array.

src/DaemonExec.ts
```diff
@@ -0,0 +1,138 @@
...
+		OpenStdin: true,
+		StdinOnce: true,
+	})
+
+	return container
+}
+
+/**
+ * Attaches input, output, and error streams to a container
+ * Default is stdin, stdout, and stderr
```
rvighne, line 94: I think this line has gone out of date, since the function has no defaults.

src/DaemonExec.ts
```diff
@@ -0,0 +1,138 @@
...
+		stderr: true,
+	})
+	const modem: ContainerModem = container.modem as ContainerModem
+	modem.demuxStream(stream, stdoutStream, stderrStream)
+	stdinStream.pipe(stream)
+	return stream
+}
+
+/**
+ * Stops the container
```
rvighne, line 121: Just curious - are we able to stop a container without removing it, and then re-use it?

src/docker-demo.ts
```diff
@@ -0,0 +1,144 @@
...
+		console.log(`Fetching image: ${image}`)
+		const stream = await importImage(docker, image)
+		const modem: DockerModem = docker.modem as DockerModem
+		modem.followProgress(stream, onFinished, onProgress)
+	} else {
+		// if image is fetched, just run command
+		onFinished(undefined).catch((err) => {
+			if (err) console.log(err)
+		})
+	}
```
rvighne, line 123: This functionality seems super useful to our daemon and also too low-level for the daemon to be concerned about. Can you move this into the library?

rvighne (changes requested):
Looks good overall. The requested changes are mostly about moving code to more useful places.

src/DaemonExec.ts
alexzma: The interfaces are different object types. Making a type definition file for all the modems would be nice, though it might take a while, and probably isn't necessary for our project.

src/DaemonExec.ts
alexzma: I believe so.

src/docker-demo.ts
alexzma: So you want me to make a function that checks if the image exists, pulls it, and then calls given callbacks through the modem?

src/DaemonExec.ts
rvighne: Oh yeah, my bad, somehow I didn't realize they have different names 😅, resolving this

src/docker-demo.ts
rvighne: Yeah exactly, the function should pull an image if it doesn't exist and resolve when this is done. That way the caller can then definitely create a container for that image without worrying about the details.

jest.config.json
```diff
@@ -0,0 +1,4 @@
+{
+	"preset": "ts-jest",
+	"testEnvironment": "node"
+}
```
Rohanator9000, line 4: Did you purposefully omit `coverageProvider`?

.eslintrc.json
```diff
@@ -1,17 +1,24 @@
...
 	"parser": "@typescript-eslint/parser",
+	"parserOptions": {
+		"project": "tsconfig.eslint.json"
+	},
 	"plugins": ["@typescript-eslint", "jest", "tsdoc"],
 	"extends": [
 		"eslint:recommended",
 		"plugin:@typescript-eslint/recommended",
+		"plugin:@typescript-eslint/recommended-requiring-type-checking",
+		"airbnb-typescript/base",
```
Rohanator9000, line 12: Why add this?

src/DaemonExec.ts
```diff
@@ -0,0 +1,190 @@
...
+}
+
+/**
+ * Returns information about the images
+ * @param docker - The docker daemon to pull information from
+ * @returns an array of ImageInfo
+ */
+export async function listImages(docker: Docker): Promise<Docker.ImageInfo[]> {
+	const images = await docker.listImages({})
+	return images
```
Rohanator9000, line 36: why not just `return docker.listImages({})`?

src/DaemonExec.ts
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * Returns information about the containers
+ * @param docker - The docker daemon to pull information from
+ * @returns an array of ContainerInfo
+ */
+export async function listContainers(
+	docker: Docker,
+): Promise<Docker.ContainerInfo[]> {
+	const containers = await docker.listContainers()
+	return containers
```
Rohanator9000, line 48: Same here - why not just `return docker.listContainers()`?

src/DaemonExec.ts
```diff
@@ -0,0 +1,190 @@
...
+ * @param name - The name of the image to pull
+ * @returns a ReadableStream to track the progress of the import
+ */
+export async function importImage(
+	docker: Docker,
+	name: string,
+): Promise<NodeJS.ReadWriteStream> {
+	const stream: NodeJS.ReadWriteStream = await (docker.pull(
+		name,
+	) as Promise<NodeJS.ReadWriteStream>)
```
Rohanator9000, line 63: Why do you need to use `as` here?

src/ZeroconfClient.ts
```diff
@@ -13,6 +13,6 @@ import { Bonjour, Browser } from 'bonjour'
 export function supplyClient(bonjour: Bonjour, client: Client): Browser {
 	const browser = bonjour.find({ type: SERVICE_TYPE })
 	browser.on('up', (service) => client.introduce(service.host, service.port))
-	client.once('done', browser.stop)
+	client.once('done', () => browser.stop())
```
Rohanator9000, line 16: Why create a new function? Why not just use `browser.stop` or `browser.stop.bind(browser)`?

src/ZeroconfDaemon.ts
```diff
@@ -30,6 +30,6 @@ export async function publishServer(
 	})
 
 	await once(service, 'up')
-	server.once('close', service.stop)
+	server.once('close', () => service.stop())
```
Rohanator9000, line 33: Same here - why not just use `service.stop` or `service.stop.bind(service)`?

src/DaemonExec.ts
alexzma: In the definition for docker.pull, it returns an any type, so we have to use "as" to assign the output to a NodeJS.ReadWriteStream.

src/DaemonExec.ts
alexzma: It was a legacy from debugging while making the code.

jest.config.json
rvighne: Yeah, it didn't seem important at this point, we can add it later though if needed.

.eslintrc.json
rvighne: The first one enables the rules that require type checking; they're disabled by default because it takes extra setup, but really it's supposed to be included. Practically, it mostly prevents you from using `any` in bad ways.

src/ZeroconfClient.ts
rvighne: They're banned by the style guide 😆. I don't mind either way, good to have it consistent though.

src/DaemonExec.ts
```diff
@@ -0,0 +1,188 @@
...
+ * Imports an image if it does not exist and then calls callbacks for progress and finish events
+ * @param docker - The docker daemon with which the image will be stored
+ * @param name - The name of the image to pull
+ * @param onFinished - a callback for when the image import is finished
+ * @param onProgress - a callback for when the progress of the image import updates
+ */
+export async function callbackImportImage(
+	docker: Docker,
+	name: string,
+	onFinished: (err?: Error) => Promise<void>,
```
rvighne, line 75: Instead of an onFinished callback, just resolve the promise when the process is done.

src/DaemonExec.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+ * Imports an image if it does not exist and then calls callbacks for progress and finish events
+ * @param docker - The docker daemon with which the image will be stored
+ * @param name - The name of the image to pull
+ * @param onFinished - a callback for when the image import is finished
+ * @param onProgress - a callback for when the progress of the image import updates
+ */
+export async function callbackImportImage(
```
rvighne, line 72: Remove "callback" from the name, maybe something like `ensureImageLoaded`.

src/DaemonExec.ts
```diff
@@ -0,0 +1,188 @@
...
+				e.RepoTags.findIndex((el: string) => el === name) > -1,
+		) < 0
+	) {
+		// if image is not fetched, fetch it (search only works if version is in image)
+		const stream = await importImage(docker, name)
+		const modem: DockerModem = docker.modem as DockerModem
+		modem.followProgress(stream, onFinished, onProgress)
+	} else {
+		// if image is fetched, just run command
+		onFinished(undefined).catch((err) => {
```
rvighne, line 97: Remove the `catch`; this function shouldn't have any say over how the caller handles errors.

src/DaemonExec.ts
```diff
@@ -0,0 +1,188 @@
...
+ * @param docker - The docker daemon with which the image will be stored
+ * @param name - The name of the image to pull
+ * @param onFinished - a callback for when the image import is finished
+ * @param onProgress - a callback for when the progress of the image import updates
+ */
+export async function callbackImportImage(
+	docker: Docker,
+	name: string,
+	onFinished: (err?: Error) => Promise<void>,
+	onProgress: (event: {
```
rvighne, line 76: I think this should be optional, since the caller might not care about the details and just wants to wait for it to be ready.

src/DaemonExec.ts
```diff
@@ -0,0 +1,188 @@
...
+		status: string
+		progressDetail: { current: number; total: number }
+		progress: string
+		id: string
+	}) => void,
+): Promise<void> {
+	const images = await listImages(docker)
+
+	if (
+		images.findIndex(
```
rvighne, line 86: `Array.some` is better than `findIndex` in this case since you don't really care about the index.

src/DaemonExec.ts
```diff
@@ -0,0 +1,188 @@
...
+			(e: Docker.ImageInfo) =>
+				e.RepoTags.findIndex((el: string) => el === name) > -1,
```
rvighne, line 88: Same suggestion here.

rvighne (changes requested):

src/DaemonExec.ts
alexzma: The onFinished callback is needed, because dockerode only says when the image is done fetching through a callback.

rvighne: Instead of making this function `async`, you can return a Promise instead. That will let you resolve the promise by just calling a continuation, which would work well here.

rvighne (approved):

PR #38, opened 2021-02-03
Unit test suite for Zeroconf

rvighne:
This adds unit testing (with mocking) for `ZeroconfClient` and `ZeroconfDaemon`.

In retrospect, unit testing these files was probably overkill, and it would be fine to just do the integration tests, but the more testing the merrier 😅.

Coming up next: integration tests.


Rohanator9000:
I think I'd rather review this once #37 is merged, otherwise I have to put in the extra work of remembering whether each change was included in #37.
(Unless there's a way to diff this against split-iface-impl in 'Files Changed').

rvighne:
> (Unless there's a way to diff this against split-iface-impl in 'Files Changed').

```shell
git checkout zc-unit-test
git diff split-iface-impl
```

Not from the UI though; Github has long ignored the feature request for dependent PR's (probably because that's not a good practice 😅)

alexzma (approved):
This looks good. jest-mock-extended seems to allow a lot more functionality than the base jest when testing classes.

rvighne:
@alexzma Yeah it's really cool, it basically generates a whole mock object from the type declaration (the interface).

src/controller.ts
```diff
@@ -9,7 +9,7 @@ import { Http2Client } from './Http2Client'
 
 const zeroconf = bonjour()
 
-function done() {
```
yashlala, line 12: I don't think the changes to this file should go in an unrelated unit test...but we're not in a place to be nitpicky so LOL go ahead

yashlala (approved):

src/controller.ts
rvighne: Good point 😅, btw this change was needed to satisfy a linter warning.

PR #39, opened 2021-02-03
Implement usecase UML diagram.

yashlala:
I accidentally pushed this change to the main uml branch instead of on a
development branch. After reverting the change to the main branch, I'm
now re-submitting this for review. 

Very self explanatory, this comment has more text than the entire PR ;)

PR #43, opened 2021-02-04
Correct UML usecase diagram.

yashlala:
Reverts cs130-w21/1#39


erlstanleylee:
Added my suggestions.

yashlala:
A bit confused -- why is sequence diagram code here? 

My intention was to quickly undo the premature merge from #39 -- all other branches would remain unchanged. 
Why is code being put in this?

erlstanleylee:
Sorry was not thinking will place in appropriate locations.

erlstanleylee (approved):
Looks good

rvighne:
@yashlala Can you explain what's going on with this PR lol. Should it be cut down to its original scope (or has that already been done) and then merged? Or discarded since it's been a while and other PR's have proceeded without it?

erlstanleylee:
This PR is soley for the use case diagram. I messed up and started adding other diagrams here, but that has been removed. The only file being reviewed is the UseCase diagram.

yashlala:
@rvighne I erroneously pushed an incorrect version of the diagram to the uml-diagrams branch. I then wanted to undo that change, so I made a quick PR with a revert message. My plan was that development would continue as usual, with this PR exactly undoing the previous one. 

Erl added his changes to the usecase diagrams to this PR directly, interpreting this branch as a "fix the previous PR" branch. This view made sense, so we rolled with it from there. 

I'll change the PR title to be more descriptive.

rvighne:
Current state: usecase

rvighne:
The separate DevOwn and Device Owner looks like a bug. Also, the diagram overall looks a bit bare. I realize we're different from the more traditional projects, but there is stuff we can add.

For example
- we can have internal use cases like "start docker daemon" that have an includes relationship to one of these.
- something like "discover daemons" and "announce daemon" can have an extends relationship to the respective use case.

erlstanleylee:
I had initially put that use case of broadcasting in there. However, I was under the impression that the device owner does not really "initiate" the broadcasting, but when the daemon is started it would automatically broadcast. Is this not the case? 

It also looks a bit bare because we removed the use case of monitoring progress from both the developer and device owner.

rvighne:
> I was under the impression that the device owner does not really "initiate" the broadcasting, but when the daemon is started it would automatically broadcast. Is this not the case?

This is true. That's why the broadcasting would be an internal use case (not directly invoked by any actor), instead it extends another use case that the user does trigger.

The meaning of this would be like "this is something the user cares about but it happens automagically when they start their daemon, they have no control over it".

Also, extends instead of includes, to show that it's like a "plugin" to the daemon's functionality.

yashlala:
Treating a network protocol as an extension of setting up the connection sounds like a slippery slope to me -- I'm not sure I agree with the "plugin" logic either (our daemon isn't customizable; changing the source code to add different functionality does not constitute a plugin, it just constitutes programming). 

That said, I do agree on broadening the scope of use cases. We should have some indication that we're joining/leaving the Junknet network (even if the user isn't sure what exactly that network consists of). 

I've added a change to the PR: let me know what you guys think. 

image

rvighne:
@yashlala Yup, this is pretty much what I meant. Also, on the client side, discovering daemons from the network is indeed an optional feature (since you could just specify hostnames instead) so that's why I recommend it be "extends" not "includes". I agree about the join/leave for the daemon though.

If you're concerned about using too many specifics in the use case diagram (e.g. Docker), just replace it with more general wording. The point remains that those are things that Junknet does, and are important to the user, even though they don't manually trigger it. Such is the case when we're trying to have a zero-conf system.

yashlala:
Added Job Monitoring to the diagram. 

If either @alexzma or @rvighne can sign off on this, then I believe we're done. 

image

alexzma (approved):
Looks good to me.

PR #44, opened 2021-02-05
Activity diagram

alexzma:
Started the activity diagram. Has overall interactions between controller, client, and daemon.


rvighne:
UML Activity diagram

rvighne (changes requested):
This is solid diagram, nice work.

Can we show the parallelism in the client somehow? Since activity diagram has the fork/join thing, we can show daemons being discovered in parallel with work being done. The join would happen once all jobs are done. (I probably haven't described that exactly how it maps to UML, but this general idea).

alexzma:
Activity-Diagram
New version of overall activity diagram.

rvighne (approved):
This is some top-tier UML craftsmanship here

yashlala (approved):

PR #45, opened 2021-02-08
JobOrderer StateMachine

Rohanator9000:


Rohanator9000:
Here's what it looks like right now: 
image

rvighne:
This is pretty good, but you might want to model this more closely after the JobOrderer implementation. E.g. add state machine variables for "inProgress", "sources", etc. and have the "job reported completed/failed" transitions update those variables. You can also use it to detect "all jobs are done".

erlstanleylee (approved):
This looks good.

Rohanator9000:
> This is pretty good, but you might want to model this more closely after the JobOrderer implementation. E.g. add state machine variables for "inProgress", "sources", etc. and have the "job reported completed/failed" transitions update those variables. You can also use it to detect "all jobs are done".

I added `sources`, `nonSources`, and `inProgress`. I omitted `jobToDependents` and `jobToNumCompletedPrereqs` because we're low on space. Regardless, the diagram is pretty wide now 😅:

JobOrdererStateMachine

rvighne (approved):
😂 yeah it might help to put some line breaks in there, but content-wise this is good.

Rohanator9000:
> 😂 yeah it might help to put some line breaks in there, but content-wise this is good.

@rvighne Added line breaks. New diagram:

image

rvighne (approved):

PR #46, opened 2021-02-10
Testing dockerode functions

alexzma:
Creates tests for the DaemonExec file. This also adds docker to travis.


test/DaemonExec.test.ts
```diff
@@ -0,0 +1,103 @@
...
+			volume,
+		)
+		expect(container).toBeDefined()
+		expect(container.id).toBeDefined()
+
+		const directory = mkdtempSync(join(tmpdir(), 'test'))
+		const target = resolve(directory, 'package.json')
+		const stdoutStream: NodeJS.WritableStream = createWriteStream(target)
+
+		await attachStreams(container, process.stdin, stdoutStream, process.stderr)
```
rvighne, line 70: Why attach the Jest process's stdin and stdout streams to this container? Also, testing stdin and stdout in addition to stdout would make it a more thorough test.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,103 @@
...
+			docker,
+			'ubuntu:18.04',
+			['/bin/ls', '/test'],
+			volume,
+		)
+		expect(container).toBeDefined()
+		expect(container.id).toBeDefined()
+		await removeContainer(container)
+	})
+	it('runs a specified container', async () => {
```
rvighne, line 52: More descriptive name for this test, maybe "attaches streams, runs a container, and reports exit status". That way the reader knows what's supposed to happen in this large test without having to understand all of it.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,103 @@
...
+			['/bin/cat', '/dev/urandom'],
+			[],
+		)
+		await container.start()
+		const containers: Docker.ContainerInfo[] = await listContainers(docker)
+		expect(
+			containers.some((cont: Docker.ContainerInfo) => cont.Id === container.id),
+		).toBeTruthy()
+		await stopContainer(container)
+		const containerInfos: Docker.ContainerInfo[] = await listContainers(docker)
```
rvighne, line 95: `containers`/`containerInfos` isn't a great pair of names, since they're the same type but sound like they're different. Maybe `containersBefore` and `containersAfter`?

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,103 @@
...
+		const container: Docker.Container = await createContainer(
+			docker,
+			'ubuntu:18.04',
+			['/bin/cat', '/test/package.json'],
+			volume,
+		)
+		expect(container).toBeDefined()
+		expect(container.id).toBeDefined()
+
+		const directory = mkdtempSync(join(tmpdir(), 'test'))
```
rvighne, line 66: Why use the `Sync` version of this method?

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,103 @@
...
+			{ fromPath: process.cwd(), toPath: '/test' },
+		]
+		const container: Docker.Container = await createContainer(
+			docker,
+			'ubuntu:18.04',
+			['/bin/ls', '/test'],
+			volume,
+		)
+		expect(container).toBeDefined()
+		expect(container.id).toBeDefined()
```
rvighne, line 49: Wouldn't this always be defined? If the container creation failed it would reject the promise (so the line before would throw an Error).

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,103 @@
...
+		await container.start()
+		const containers: Docker.ContainerInfo[] = await listContainers(docker)
+		expect(
+			containers.some((cont: Docker.ContainerInfo) => cont.Id === container.id),
+		).toBeTruthy()
+		await stopContainer(container)
+		const containerInfos: Docker.ContainerInfo[] = await listContainers(docker)
+		expect(
+			containerInfos.some(
+				(cont: Docker.ContainerInfo) => cont.Id === container.id,
```
rvighne, line 98: Remove the type annotation (also from the previous instance). It makes it harder to read, plus we already know the item type from the array type.

rvighne (changes requested):
Great work. Now that we have this automated test, can we remove the "manual" test in `bin/docker-demo.ts`? Or is there still some functionality tested there and not here?

Also, we need a way to run integration tests (like this one) separately from unit tests. Integration tests take a long time and have external dependencies like Docker, which some of us (me) don't have installed. I'll try to send a PR soon that adds that functionality.

alexzma:
> Great work. Now that we have this automated test, can we remove the "manual" test in `bin/docker-demo.ts`? Or is there still some functionality tested there and not here?

I honestly thought docker-demo.ts was already deleted. I'll delete that.

test/DaemonExec.test.ts
alexzma: Jest gives a warning if there are no explicit tests in an it block. Also, id's are needed for container manipulation, so if the container doesn't have an id, the other tests won't work.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,103 @@
...
+interface ExpectedStatus {
+	Error: unknown
+	StatusCode: number
+}
+
+const docker = new Docker()
+jest.setTimeout(300000)
+
+describe('DaemonExec', () => {
+	it('pulls a specified image', async () => {
```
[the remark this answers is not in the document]
alexzma: All tests within a describe block happen in sequential order.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+		await ensureImageImport(docker, testImage)
+		const newInfos = await listImages(docker)
+
+		expect(
+			newInfos.some((info) => info.RepoTags.some((tag) => tag === testImage)),
+		).toBeTruthy()
+	}, 60000)
+
+	it('pulls an existing image', async () => {
```
rvighne, line 55: How do you know this actually quits early with the existing image as opposed to downloading it? Maybe Travis has a fast internet connection ;)

Add a comment to document your assumption that a new image can't get downloaded in less than 5 seconds.

---

Btw, don't actually spend time on this since we need to focus on the SSH daemon, but it's a good idea to add a unit test for this method (passing it a fake Docker) in addition to this integration test, so you can test that it's following the right logic. Usually unit tests are written first.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+		errorFile = resolve(tempDir, 'error')
+		const stderrStream: NodeJS.WritableStream = createWriteStream(errorFile)
+
+		const stream = await attachStreams(
+			container,
+			stdinStream,
+			stdoutStream,
+			stderrStream,
+		)
+		expect(stream).toBeDefined()
```
rvighne, line 92: This assertion isn't needed right? If `attachStreams` fails, it should reject the promise rather than return `undefined` (so it'll never get to this line).

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+	})
+
+	it('attaches streams to a specified container', async () => {
+		// inputs package.json to stdin, gets stdout and stderr in temp directory
+
+		tempDir = await fs.mkdtemp(join(tmpdir(), 'test'))
+
+		const stdinStream: NodeJS.ReadableStream = createReadStream(inputFile)
+
+		outputFile = resolve(tempDir, inputFile)
```
rvighne, line 80: Why is it outputting to `package.json`?

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+		const data: ExpectedStatus = await (container.wait() as Promise<ExpectedStatus>)
+
+		expect(data.Error).toBeNull()
+		expect(data.StatusCode).toBe(0)
+	})
+
+	it('writes the correct data to the streams', async () => {
+		const original = await fs.readFile(inputFile)
+		const copy = await fs.readFile(outputFile)
+		expect(original).toEqual(copy)
```
rvighne, line 107: The [jest docs for toEqual](https://jestjs.io/docs/en/expect#toequalvalue) seem to suggest that we should use the Buffer `equals` method directly.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+		const error = await fs.readFile(errorFile)
+		expect(Buffer.byteLength(error)).toEqual(0)
```
rvighne, line 110: Why not `error.length`?

Also, `toEqual` is if you want to recursively compare objects; you probably want to use `toBe(0)` here.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+			volumes,
+		)
+
+		await volumeContainer.start()
+		await volumeContainer.wait()
+
+		// volumes should have the correct files copied over
+		const original = await fs.readFile(inputFile)
+		const copy = await fs.readFile(resolve(tempDir, inputFile))
+		expect(original).toEqual(copy)
```
rvighne, line 138: Same comment as above about comparing Buffer.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+	it('removes a specified container', async () => {
+		await removeContainer(container)
+		await expect(container.start()).rejects.toThrow()
+	})
+
+	it('creates and runs containers with volumes and gives correct outputs', async () => {
+		// make volumes from current directory and temporary directory
+		const volumes: VolumeDefinition[] = [
+			{ fromPath: process.cwd(), toPath: '/test' },
+			{ fromPath: tempDir, toPath: '/test2' },
```
rvighne, line 122: Better names for `toPath`.

test/DaemonExec.test.ts
```diff
@@ -0,0 +1,167 @@
...
+		const error = await fs.readFile(errorFile)
+		expect(Buffer.byteLength(error)).toEqual(0)
+	})
+
+	it('removes a specified container', async () => {
+		await removeContainer(container)
+		await expect(container.start()).rejects.toThrow()
+	})
+
+	it('creates and runs containers with volumes and gives correct outputs', async () => {
```
rvighne, line 118: Nice, I like how you separated the volume test from streams and exit codes.

rvighne (changes requested):
Looks good overall. There is a potential bug in the "attaches streams to a specified container" testcase.

test/DaemonExec.test.ts
alexzma: This test is for the behavior when a requested image already exists. It is not necessarily testing whether it will download the image again, but whether the image exists after the method is called, and the method does not throw an error. It is a convenient coincidence that the image does not get downloaded.

test/DaemonExec.test.ts
alexzma: There is a warning if there is no expect in the test. I would like to avoid warnings if possible.

test/DaemonExec.test.ts
alexzma: It's not outputting to package.json. It is outputting to a file in a temporary directory that happens to be named package.json.

test/DaemonExec.test.ts
rvighne: I see, this makes sense then.

test/DaemonExec.test.ts
rvighne: 🤦‍♂️right. I need to read more carefully :)

test/DaemonExec.test.ts
rvighne: Fair enough. Could you add a comment explaining that's why you have the assertion?

rvighne:
Forgot to mention: can you move the test file under `test/integ` instead of `test`? That will be compatible with #65.

test/DaemonExec.test.ts
alexzma: length gives the size of the Buffer. byteLength gives the size of the contents.

rvighne: Is there a difference?

rvighne (approved):

test/DaemonExec.test.ts
alexzma: Probably not in this case, but if you create a buffer of a certain size, and don't fill it with any data, then length returns the size, and byteLength returns 0.

rvighne:
Could you make a tarball of the `coverage` directory under your project dir (after running integration tests on this branch) and send it to me? The Travis build log is showing some weird output for DaemonExec coverage and I'd like to try to debug it.

pjot726 (approved):
Looks good to me, the testcases are pretty comprehensive.

PR #47, opened 2021-02-10
Class diagram.

alexzma:
This is the class diagram.

### DIFF
--- a/Class/class.plantuml
+++ b/Class/class.plantuml
@@ -39,7 +39,7 @@ class HeapJobOrderer implements JobOrderer{
 + constructor(in rootJobs: Job[])
 }
 
-JobOrderer o-- Job
+JobOrderer "1" o-- "*" Job
 Client "1" --o "1" Controller
 
 interface Client{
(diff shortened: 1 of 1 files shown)


rvighne:
Rendering of the diagram: Junknet UML class diagram preview

In the future, let's always attach this link to the PR so others can easily see the diagram.

rvighne (approved):
Overall looks good. I would recommend
- adding multiplicity to all of the "aggregation arrows" (the ones with hollow diamond)
- make the Docker-related functions private, and remove some of them and simplify the signatures (that box takes up way too much of the diagram and we might not be able to paste this into a google doc at that size)

alexzma:
ClassDiagram
Updated class diagram

rvighne:
Forgot to mention this before: client to daemon multiplicity should be many-to-many.

Also, the Daemon function signatures are still not really a manageable size. We don't have to be completely accurate with the types there, maybe just `attachStreams(in stdin, out stdout, out stderr)`?

And we can remove `in` from all parameters, since that's the default. Any other space saving ideas?

yashlala:
The Parser code is currently in a class; however, it's entirely static (might as well just put it in a module). 
I'm not sure how this code would fit into a class diagram, as the Controller wouldn't technically contain any Parser objects. Any objection to my adding the below code to the diagram? 

```
class Parser{
+ parseMakefile(in file: string): Job [*]
}
Controller "1" --* "1" Parser
```

alexzma:
> The Parser code is currently in a class; however, it's entirely static (might as well just put it in a module).
> I'm not sure how this code would fit into a class diagram, as the Controller wouldn't technically contain any Parser objects. I think it might be safer to add it to the diagram anyways -- any objections to my adding something the below snippet?
> 
> ```
> class Parser{
> + parseMakefile(in file: string): Job [*]
> }
> Controller "1" --* "1" Parser
> ```

No objections

yashlala:
Alright. Added it to the PR, with some minor tweaks (composition arrow was in the wrong direction). 

Diagram shown below. 

image

yashlala (approved):

PR #48, opened 2021-02-11
sequence diagram(sequence1.plantuml)

erlstanleylee:
Added my sequence diagram focusing on the communication that client does with joborderer and daemon.

### DIFF
--- a/Sequence/Sequence1.plantuml
+++ b/Sequence/Sequence1.plantuml
@@ -1,0 +1,74 @@
+@startuml
+
+/'
+This sequence diagram was created to emphasize that the client did not have to wait for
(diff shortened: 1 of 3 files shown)


yashlala (changes requested):
This seems to indicate that the user talks directly to the client instead of going through the controller. Sending the user request through the controller first would be consistent with the rest of the documentation.

erlstanleylee:
@yashlala I added the controller and parser entity.

rvighne:
# Failure Sequence
UML Sequence Diagram for Failure

# Success Sequence
UML Sequence diagram for Success

yashlala (approved):

rvighne (approved):
Looks good.

PR #49, opened 2021-02-11
Added daemon state machine.

erlstanleylee:
Added state-machine for daemon.

### DIFF
--- a/StateMachine/daemonStateMachine.plantuml
+++ b/StateMachine/daemonStateMachine.plantuml
@@ -1,0 +1,29 @@
+@startuml
+
+
+state HasJob {
+  state "Do Work" as Work
+  state "Troubleshoot with Client" as Trouble
+  state "Return Output to Client" as Output
+
+  [*] --> Work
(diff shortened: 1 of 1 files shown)


alexzma (changes requested):
I think client cancel job should go from HasJob to Advertising instead of from Do Work to Advertising, since the client could cancel while the daemon is sending back data or troubleshooting. Then, the daemon should stop sending the data to the client concerning the job.

Also, could you label the transition between return output and advertising and the transition between clean up and server closed? I know they are pretty self-explanatory, but just in case.

erlstanleylee:
@alexzma Made the requested changes.

alexzma (approved):
Looks good!

rvighne:
UML State Machine diagram for Daemon

rvighne (approved):
Can you remove the final "Close Server" state (i.e.  make it go directly to the end state)? Other than that, LGTM.

Btw, the "Advertising" phase of isn't exactly how it works now, but this is still a fine diagram.

PR #60, opened 2021-02-12
CLI Controller

yashlala:
Addresses issue #7.


yashlala:
I'm a bit confused -- `npm test:local` works just fine on this branch, all tests pass, and ESlint doesn't complain. The functionality is there. 
However, it doesn't compile when we try to merge to master, failing with this err message: 

```
> tsc

src/Controller/CommandLineController.ts:17:20 - error TS2349: This expression is not callable.

  Type '{ default: Argv<{}>; alias<K1 extends never, K2 extends string>(shortName: K1, longName: K2 | readonly K2[]): Argv<{ [key in K2]: {}[K1]; }>; alias<K1 extends never, K2 extends string>(shortName: K2, longName: K1 | readonly K1[]): Argv<...>; alias(shortName: string | readonly string[], longName: string | readonly st...' has no call signatures.

17  const yargsArgv = yargs(argv.slice(2))

                      ~~~~~

  src/Controller/CommandLineController.ts:1:1

    1 import * as yargs from 'yargs'

      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

    Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.
```

I'm changing around the imports as per the error message, but can't seem to get it to compile in both this branch and in master. Continuing to work on it.

yashlala:
OK, I copied all of my code to a fresh branch based off master and it seems to work. This branch probably contained some unrelated config change that was causing the error. 
Closing; will resubmit PR with new branch.

PR #61, opened 2021-02-13
Restructure Directories

Rohanator9000:
Feel free to suggest otherwise if you dislike this.


rvighne:
@Rohanator9000 mind if I make one change to this? I want to move the programs into a `bin` directory under the root (as opposed to `src/programs`) and change some related configs.

That way we can exclude it from the docs generator and unittest coverage and further enforce not mixing the two.

Rohanator9000:
> @Rohanator9000 mind if I make one change to this? I want to move the programs into a `bin` directory under the root (as opposed to `src/programs`) and change some related configs.
> 
> 
> 
> That way we can exclude it from the docs generator and unittest coverage and further enforce not mixing the two.

@rvighne sure no problem

rvighne (approved):

Rohanator9000:
@rvighne LGTM. Need someone else to review so I can merge.

yashlala (approved):

PR #65, opened 2021-02-19
Multiple test suites and small changes for Part C

rvighne:
This is a bit of a mixed-bag PR of changes that are important for devops and Part C. It sets the version number to 0.1.0, so we can make a release once this is merged. Changes in decreasing order of significance:
1. Split the test suite into `unit` and `integ` using Jest projects. This is important because we're about to add integration tests, which have external dependencies and should be run in CI but not every developer's local machine; see [this comment](https://github.com/cs130-w21/1/pull/46#pullrequestreview-592634916).
2. Show code coverage on all test runs. It's not enforced, but hopefully seeing it will encourage you to get to 100% 😊
3. Move the Developer's Guide from CONTRIBUTING.md into README.md and clarify some things. This is a Part C requirement.
4. Fix an issue where the `index.ts` was shown in the typedoc site, which duplicates the other docs. This is tangentially related to Part C.
5. Update dependencies.


package.json
```diff
@@ -23,7 +23,8 @@
     "postfix": "prettier -w .",
     "pretest": "eslint .",
     "test": "jest",
-    "posttest": "prettier -c ."
+    "posttest": "prettier -c .",
+    "test:dev": "npm run test -- --runInBand --selectProjects unit"
```
Rohanator9000, line 27: I feel like "dev" is kind of misleading because most of the time when I'm developing, I will run all tests before committing.

In fact, why include it at all? Because integration tests can take a long time?

rvighne: The motivation for this change is that actually you don't want to run all tests before committing. Always run unit tests, but integration tests have external dependencies which you shouldn't have to install, like Docker (and yeah they take a long time as well).

This change makes it so you can run unit tests separately from the others, while CI (or just `npm test`, like before) runs everything.

typedoc.json
```diff
@@ -1,6 +1,6 @@
 {
 	"entryPoints": "src",
+	"exclude": "src/index.ts",
 	"includeVersion": true,
-	"readme": "CONTRIBUTING.md",
```
alexzma, line 4: Shouldn't we replace the readme field with README.md, since it does what CONTRIBUTING.md used to do? It's the only document that explains how to run our program currently.

alexzma (changes requested):
Looks good overall.

typedoc.json
rvighne: Yes, `README.md` is the default. It will appear in the front page of the docs site.

alexzma (approved):
Looks good to me.

package.json
rvighne: Updated the name to `test:local`.

Rohanator9000: Why not `test:unit`? Describe it exactly.

rvighne: It doesn't just run unit tests (it does linting, unit tests, then formatting) so `test:unit` wouldn't describe it exactly. Though I'm open to other suggestions.

Rohanator9000 (approved):

PR #69, opened 2021-02-23
Streamline the dev process

rvighne:
This add several small changes to improve the testing and development process. The most important change is that the compiler is now run during `test` to do type-checking (this was already enforced in CI, but didn't run locally unless you asked for it).

We may add more improvements following the meeting on Thursday.


yashlala (approved):
We'll have to disable some unicorn rules on a case-by-case basis later (eg. `no-null`), but this is a good starting point.

rvighne:
> We'll have to disable some unicorn rules on a case-by-case basis later (eg. `no-null`), but this is a good starting point.

The only rule we're using from unicorn is `no-array-for-each` (note we're not extending from unicorn, just adding one under the `rules` section).

.eslintrc.json
```diff
@@ -13,28 +13,26 @@
 		"plugin:promise/recommended",
 		"plugin:jest/recommended",
 		"plugin:jest/style",
-		"prettier",
-		"prettier/@typescript-eslint"
```
Rohanator9000, line 17: Why did you remove this?

.eslintrc.json
```diff
@@ -13,28 +13,26 @@
...
-		"tsdoc/syntax": "error",
 		"no-restricted-syntax": [
 			"error",
 			"ForInStatement",
 			"LabeledStatement",
 			"WithStatement"
 		],
 		"no-return-assign": ["error", "except-parens"],
-		"github/array-foreach": ["error"],
+		"unicorn/no-array-for-each": "error",
```
Rohanator9000, line 28: Why did you switch from GitHub to unicorn?

package.json
```diff
@@ -1,29 +1,32 @@
 {
   "name": "junknet",
   "version": "0.1.0",
-  "description": "Zero-config distributed parallel processing",
+  "description": "Zero-config distributed parallel processing.",
   "license": "Apache-2.0",
   "files": [
     "dist",
-    "docs"
+    "!**/*.tsbuildinfo"
```
Rohanator9000, line 8: We're not using the `--incremental` flag, right? What else would produce `.tsbuildinfo`?

.gitignore
```diff
@@ -1,9 +1,18 @@
 node_modules
+*.tgz
```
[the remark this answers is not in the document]
rvighne: npm pack

.eslintrc.json
rvighne: Upgraded `esling-config-prettier`, turns out they have removed the second config because it's built in to the first.

.eslintrc.json
rvighne: See the commit log. This is the same type of rule as `github/array-foreach`, but it's auto-fixable.

Also, unicorn has a lot of other interesting rules we may want to use in the future (not enabled now cause I don't want to break existing stuff).

package.json
rvighne: `composite: true` implies `incremental: true`

Rohanator9000 (approved):

PR #70, opened 2021-02-25
Static make binary compilation

alexzma:
This branch should take steps that will end up with a static binary for make.


rvighne:
Closing so it doesn't appear as unfinished work. When this feature comes back into scope, we can reopen this PR.

PR #71, opened 2021-02-26
Implement Command Line Interface

### LINKED ISSUE #7: Command line user input Interface
## User Story
As a developer, I would like to specify a Makefile and a Docker image so that the work can be sent to the worker machines.

## Detailed Description

As a user, I would like to be able to interact with the controller, and specify the work I would like done. 
I should be able to do so via a standard UNIX shell interface; ie. by passing options to the frontend. 

## Tasks

- [x] Create command line front end
- [ ] Pass the information to the rest of the controller. 
- [x] Write tests. 

## Acceptance Criteria

- [x] Given a command line invocation with options, store these arguments. 

# Estimated points
1

yashlala:
Implements and tests the Command Line Interface, closes #7. 

This pull request serves as a "clean start" for the code formerly in issue #60 -- a beautiful new world where possibilities are endless, commits are atomic, and all tests pass.


package-lock.json
```diff
@@ -1,9736 +1,8 @@
 {
   "name": "junknet",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
rvighne, line 4: You're running an old version of `npm` and it's trying to downgrade the lockfile (more importantly, it doesn't understand the lockfile so you may not get the right versions).

    npm i -g npm

should update it.

package.json
```diff
@@ -53,6 +52,7 @@
     "typescript": "^4.1.3"
   },
   "dependencies": {
+    "@types/yargs": "^16.0.0",
```
rvighne, line 55: `@types/yargs` should stay in devDependencies, not dependencies. You might have done `npm i` when you meant to do `npm i -D`. You could just checkout this and `package-lock.json`.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
...
+	makefile: string
+	docker_image: string
+	target: string | null
+}
+
+/**
+ * Interpret a list of arguments, returning the relevant ones in the f
+ *
+ * @param argv - the invoking process's command-line arguments.
+ * @returns JunknetInterface object containing option values.
```
rvighne, line 13: Don't put the return type name in the docstring since it's redundant (also, in this case not consistent).

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
+import yargs from 'yargs'
+
+export interface JunknetArguments {
+	makefile: string
+	docker_image: string
+	target: string | null
```
rvighne, line 6: Use `undefined` instead of `null` to indicate optional fields.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
+import yargs from 'yargs'
+
+export interface JunknetArguments {
+	makefile: string
+	docker_image: string
```
rvighne, line 5: Naming convention for JS is camelCase.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
...
+	target: string | null
+}
+
+/**
+ * Interpret a list of arguments, returning the relevant ones in the f
```
rvighne, line 10: Is this sentence incomplete?

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @param argv - the invoking process's command-line arguments.
+ * @returns JunknetInterface object containing option values.
+ */
+export function interpretArgv(argv: readonly string[]): JunknetArguments {
+	// Use only the first two elements; Node.js appends extra elements to process.argv.
+	const yargsArgv = yargs(argv.slice(2)).options({
```
rvighne, line 17: It would be better to assume the caller has provided what we actually want to parse. Say the arguments don't come from `process.argv` originally, you don't want to make them do something silly like prepend two placeholder items.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
...
+			alias: 'target',
+			type: 'string',
+			desc: 'The Makefile target to build',
+		},
+	}).argv
+
+	let target = null
+	if (yargsArgv.t !== undefined) {
+		target = yargsArgv.t
+	}
```
rvighne, line 40: const target = yargsArgv.t ?? null

or better yet, don't use null, just undefined.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
...
+		f: {
+			alias: 'makefile',
+			type: 'string',
+			default: 'Makefile',
+			desc: 'The Makefile to process',
+		},
+		i: {
+			alias: 'docker-image',
+			type: 'string',
+			default: 'ubuntu:18.04',
```
rvighne, line 27: Don't provide an arbitrary default.

Btw, this image doesn't contain `make`.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,47 @@
...
+			desc: 'The Docker Image to run',
+		},
+		t: {
+			alias: 'target',
+			type: 'string',
+			desc: 'The Makefile target to build',
+		},
```
rvighne, line 34: Shouldn't this be a positional argument, and allow dups? So you can do

    junknet-make all install clean

Instead of

    junknet-make -t all

test/unit/CommandLineController.test.ts
```diff
@@ -0,0 +1,77 @@
...
+			'0',
+			'1',
+			'2',
+			'3',
+			'--makefile',
+			'mf',
+			'--docker-image',
+			'di',
+			'--target',
+			't',
```
rvighne, line 62: Make "mf", "di", and "t" longer so you understand what "type" they are without context, e.g. "build.make", "fake-image:latest", "clean". Also, extract them into variables with names so they're not "magic strings".

rvighne (changes requested):
Looks good. Clean code as usual 😊. The requested changes are all localized to specific things, not design-level.

package-lock.json
yashlala: I was using my OS's version of `npm` (6.14.11), which didn't support lockfile v2s. 

New commit addresses this. I'm now using a locally built version, so this issue won't recur.

src/Controller/CommandLineController.ts
yashlala: I noticed that accessing nonexistent property names yielded `undefined`. 
I opted to use `null` in the union type to distinguish it from this case; most of my reasoning was based on [this](https://stackoverflow.com/questions/461966/why-is-there-a-null-value-in-javascript) post.  

Is using `undefined` in this way a JS convention?

src/Controller/CommandLineController.ts
yashlala: old habits die hard ;)

src/Controller/CommandLineController.ts
yashlala: This makes sense -- we do want our interface to resemble GNU make. 

Now that I'm thinking more about interface compatibility, we should probably also: 

- Use short-option letters that aren't currently used by GNU make.
- Pass any extraneous options to `make`, so things like `make -B` won't cause our program to explode.  
- Search the current directory for `GNUmakefile`, `Makefile`, and `makefile`, using the first file it finds. 

I'll add these to the PR before merging.

src/Controller/CommandLineController.ts
yashlala: Sure. I'll make the dockerfile the first positional argument.

src/Controller/CommandLineController.ts
rvighne: Take a look at the argument parser in [server.ts](https://github.com/cs130-w21/1/blob/master/bin/server.ts). I added a comment like this:

    // Modeled after sshd(8) where possible, see https://man.openbsd.org/sshd.

And made the options be a subset of sshd. That way in the future whoever wants to add something that's in sshd will use the same name and if it's not in sshd then they'll use a name that's not claimed.

rvighne:
Does this help? https://github.com/yargs/yargs/blob/master/docs/api.md#failfn--boolean

> Method to execute when a failure occurs, rather than printing the failure message.
>
> Providing false as a value for fn can be used to prevent yargs from exiting and printing a failure message. This is useful if you wish to handle failures yourself using try/catch and .getHelp().

yashlala:
> Does this help? https://github.com/yargs/yargs/blob/master/docs/api.md#failfn--boolean
> 
> > Method to execute when a failure occurs, rather than printing the failure message.
> > Providing false as a value for fn can be used to prevent yargs from exiting and printing a failure message. This is useful if you wish to handle failures yourself using try/catch and .getHelp().

Unfortunately, this doesn't quite cover all the cases. If we receive a `--version` or `--help` command, we want to exit (and this function isn't called in those cases, although it does handle wrong versions. 

The two options I see are: 

1. Ask `jest` to override `process.exit()` with mock functions. I've worked on this, but am having difficulty passing in a lambda function that never returns (it must return `never` in order to match the signature of `process.exit()`). 
2. Stop yargs from exiting in any case (there's a simple flag to pass). I'll then test for `--version` or an incorrect positional argument, and throw an exception/return a value specifying what to terminate the program with. This still breaks encapsulation (because yargs prints to stdout regardless of what we do), but at least the `exit` code won't have side effects. 

Probably going to try option 2, because option 1 isn't working out and option 2 is simpler. 

The second option

yashlala:
@rvighne All issues should be resolved now.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,96 @@
+import yargs from 'yargs'
+
+export interface ArgvData {
```
Rohanator9000, line 3: Can you write documentation for this interface (even if its brief)? I had to read the rest of the file to figure out what it's used for.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,96 @@
...
+			dockerImage: '',
+			targets: [],
+			cleanExit: true,
+			invalidArguments: false,
+		}
+	}
+
+	const positionalArgs = yargsArgv._ as string[]
+
+	let dockerImage
```
Rohanator9000, line 67: Why not accept the docker image as an option (like f/makefile) instead of a positional arg?

test/unit/CommandLineController.test.ts
```diff
@@ -0,0 +1,99 @@
+/**
+ * Tests for the Command-Line Argument Parser
+ */
+
+import * as cli from '../../src/Controller/CommandLineController'
```
Rohanator9000, line 5: Very nitpicky, but shouldn't the import go before the test documentation? I'm actually not sure if that's the convention, but it just seems right to me haha.

Rohanator9000 (changes requested):

src/Controller/CommandLineController.ts
yashlala: We require a docker image when building (we're not supplying a default). 
It seemed like bad design to ask for an option `-i` if we know that the data is mandatory.

test/unit/CommandLineController.test.ts
yashlala: Quick nitpick, but quick fix ;)

src/Controller/CommandLineController.ts
yashlala: Fixed.

src/Controller/CommandLineController.ts
```diff
@@ -0,0 +1,119 @@
+import yargs from 'yargs'
+
+/**
+ * Contains processed argument data (usually from `process.argv`.
```
Rohanator9000, line 4: Missing parentheses :).

Rohanator9000 (changes requested):

src/Controller/CommandLineController.ts
Rohanator9000: Is makefile an option because it's optional? Would we just look for a file in the immediate directory called `Makefile`?

yashlala: Yes. We look for `GNUmakefile`, `Makefile`, and `makefile` in that order (as with GNU make).

Rohanator9000 (approved):

rvighne (approved):

PR #72, opened 2021-02-26
Added SIGINT catch for exit controller. Added log job on progress eve…

erlstanleylee:
…nt. Update visibility of of closeAllDaemons and Finish to Client interface.


bin/controller.ts
```diff
@@ -28,9 +28,20 @@ const job1: Job = new NormalJob('first', new Set([job4, job2]))
...
+})
 once(client, 'done').then(clientDone).catch(console.error)
 
+process.on('SIGINT', () => {
+	console.log('\nClosing Daemons\n')
+
+	client.closeAllDaemonsAndFinish()
+
+	console.log('Exit Process\n')
+	process.exit()
```
rvighne, line 42: It shouldn't be necessary to force-exit the process. Ideally, once you close all the sockets, you just wait for the event loop will run out. What happens if you remove this line?

src/Client/Client.ts
```diff
@@ -51,4 +51,9 @@ export interface Client extends TypedEmitter<ClientEvents> {
 	 * @param port - port number of daemon on the host
 	 */
 	introduce(host: string, port: number): void
+
+	/**
+	 * Close all daemons
+	 */
+	closeAllDaemonsAndFinish(): void
```
rvighne, line 58: Could you rename this method to be shorter in both classes? E.g. `quit`.

rvighne (changes requested):
Good work.

src/Client/Client.ts
erlstanleylee: Will do.

bin/controller.ts
erlstanleylee: Yup do not need to force-exit. Removed this line.

rvighne (approved):

bin/controller.ts
```diff
@@ -28,9 +28,19 @@ const job1: Job = new NormalJob('first', new Set([job4, job2]))
 const job5: Job = new NormalJob('fifth', new Set([job1]))
 
 const client: Client = new Http2Client(new HeapJobOrderer([job5]))
-client.on('progress', console.log)
+client.on('progress', (FinishedJob) => {
```
Rohanator9000, line 31: Why is the variable `FinishedJob` uppercase?

erlstanleylee: My mistake. Will change.

yashlala (approved):

PR #73, opened 2021-02-27
Makefile Trace Generation and Parsing

Rohanator9000:


Rohanator9000:
@rvighne @yashlala My code is not ready for actual review, but could you help me structure my code? Namely `src/Controller/MakefileToJobTree.ts` and `src/Controller/MakefileTraceToJobTree.ts`.

I initially tried to put my code in an object since that's what I've been doing up to now for this project. However, that doesn't really make sense because my code is pretty functional. Plus TS complains when any non-static object methods don't use `this`, and I don't want an object full of static methods.

Currently, I've just stuffed the code into two files, but it just seems poorly organized to me. Any recommendations?

yashlala:
In my code, I went ahead with the "class full of static methods" approach; it was the closest thing TS had to a namespace.

rvighne:
> I initially tried to put my code in an object since that's what I've been doing up to now for this project. However, that doesn't really make sense because my code is pretty functional.

I agree, and the way you've structured it is fine. Just export the functions from your module that should be publicly accessible (of course, the less the better). You don't need to use OOP when it's not helpful.

Re. @yashlala about namespaces: a module only exports names that are explicitly marked as `export`, so you don't need to create a class in order to get that functionality. Make as many top-level functions as you want since they're hidden, and `export` the ones that make up your module's public API. It's the same system as Haskell and many other FP languages.

Rohanator9000:
`test/supplementary files/makefile test 1/Makefile` kind of seems like bad file organization.

src/Controller/MakefileTraceToJobTree.ts
```diff
@@ -0,0 +1,210 @@
...
+ *
+ * Used during trace parsing.
+ */
+interface TargetLineAndCommands {
+	targetLine: string
+	commands: string[]
+}
+
+// Regexes used when interpreting target-lines.
+const targetLineRegexWithPrereqs = /^.+: update target '(.+)' due to: (.+)$/
```
yashlala, line 26: These regexes seem dangerous. Can we replace this with something like: 

```suggestion
const targetLineRegexWithPrereqs = /^.+: update target '([A-Za-z0-9_/.]+)' due to: (([A-Za-z0-9_/.]+)$/
```

src/Job/NormalJobOptions.ts
```diff
@@ -0,0 +1,18 @@
+import { Job, JobEnv } from './Job'
+
+/**
+ * An interface describing options for a {@link NormalJob}.
+ *
+ * target: The target file that the Job will produce. Must be unique between jobs in the same dependency graph.
+ * commands: An array of the commands to run to build the target.
+ * prerequisiteJobs: A set containing all of this job's prerequisite Jobs.
+ * prerequisiteFiles: A set containing all of this job's prerequisite files
```
yashlala, line 9: In what sense are we using "prerequisite job" and "prerequisite file" here? How are the 2 distinct?  Can we give a brief example to make the documentation a bit more clear? 

```suggestion
 * prerequisiteFiles: A set containing the names of all files needed by the job to run (eg. "sourceCodeToCompile.js"). 
```

bin/controller.ts
```diff
@@ -14,11 +14,23 @@ import {
...
+const job1: Job = new NormalJob({
+	target: 'first',
+	commands: [],
+	prerequisiteJobs: new Set([job4, job2]),
+})
+const job5: Job = new NormalJob({
+	target: 'fifth',
+	commands: [],
+	prerequisiteJobs: new Set([job1]),
+})
```
yashlala, line 33: What's the purpose of this code?

yashlala (changes requested):
Mad respect -- you've put a lot of time into this, and I'm grateful for it. 

I've requested some minor changes (making the regex a bit less aggressive, etc), but by and large this is solid work.

bin/controller.ts
Rohanator9000: We initially included this code to test the controller-daemon interaction. I just realized you merged the yargs code so I'll integrate that.

src/Controller/MakefileTraceToJobTree.ts
Rohanator9000: I updated the regex.

yashlala (approved):

yashlala:
@Rohanator9000 I noticed a potential bug in `bin/controller.ts`, so I made some changes to this branch (resolving my former suggestion while I was at it).

bin/controller.ts
```diff
@@ -7,40 +7,57 @@ import {
...
+// we want `--help` messages to take priority over invalid arguments.
+if (cliArgs.invalidArguments) {
+	console.error(
+		new Error(`Invalid arguments supplied: ${process.argv.slice(2)}.`),
+	)
+	process.exit(1)
+}
+
+makefileToJobTree({ filePath: cliArgs.makefile, targets: cliArgs.targets })
+	.then((rootJobs) => {
```
rvighne, line 34: Move this block into a separate function so you can make it `async`. That way you can use `await makefileToJobTree()` and make the control flow a lot easier to understand.

bin/controller.ts
```diff
@@ -7,40 +7,57 @@ import {
...
+
+		const browser = supplyClient(zeroconf, client)
+		browser.on('up', console.info)
+		browser.on('down', console.info)
+
+		return client
+	})
+	.catch((error) => {
+		console.error(error)
+		process.exit(1)
```
rvighne, line 77: Better to just set `process.exitCode = 1` and let the process terminate normally sometime soon. See an example on lines 42-43.

src/Controller/MakefileToJobTree.ts
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	if (options.targets) {
+		command += ` ${options.targets.join(' ')}`
+	}
+
+	let stdout: string
+	let stderr: string
+	try {
+		;({ stdout, stderr } = await promisify(exec)(command))
```
rvighne, line 36: Use `execFile` instead of `exec`. That will execute the process directly instead of creating a shell, and avoids all the pitfalls with escaping the commandline arguments since you pass them in as an array instead of string.

src/Controller/MakefileToJobTree.ts
```diff
@@ -0,0 +1,63 @@
...
+	filePath?: string
+	targets?: string[]
+}): Promise<string> {
+	let command = 'make --trace --dry-run'
+
+	if (options.filePath) {
+		const fileName = basename(options.filePath)
+		const directoryName = dirname(options.filePath)
+
+		command += ` -f "${fileName}" --directory="${directoryName}"`
```
rvighne, line 26: And what if I want my fileName to be `"; sudo rm -rf /`?

src/Controller/MakefileToJobTree.ts
```diff
@@ -0,0 +1,63 @@
...
+		command += ` ${options.targets.join(' ')}`
+	}
+
+	let stdout: string
+	let stderr: string
+	try {
+		;({ stdout, stderr } = await promisify(exec)(command))
+	} catch (error) {
+		throw new MakeTracingError(
+			`Error while generating make trace using command "${command}":\n${error}`,
```
rvighne, line 39: Do not put newlines in the `Error` message field. It usually makes logging output hard to read.

src/Controller/MakefileToJobTree.ts
```diff
@@ -0,0 +1,63 @@
...
+		)
+	}
+
+	if (stderr.length > 0) {
+		throw new MakeTracingError(
+			`Error while generating make trace using command "${command}".\n\nStderr:\n${stderr}\n\nStdout:${stdout}`,
```
rvighne, line 45: Couldn't `stderr` and `stdout` end up being massive strings? That would be a pretty unhelpful error message. It's better to add those as fields in the error object (and extend the ctor to accept those fields).

src/Job/NormalJob.ts
```diff
@@ -14,64 +15,110 @@ const DEFAULT_ENV: JobEnv = Object.freeze({
...
+		return this.prerequisiteJobs.size
 	}
 
 	/**
 	 * JavaScript calls this function whenever Job is casted to a string.
 	 *
-	 * @returns This job's name.
+	 * @returns A string detailing this job's target, as well as any prerequisite jobs or files.
 	 */
 	public toString(): string {
```
rvighne, line 87: Whoever has to debug code using `Job` is going to be really happy to see such detailed toString output 😀

src/Job/NormalJob.ts
```diff
@@ -14,64 +15,110 @@ const DEFAULT_ENV: JobEnv = Object.freeze({
...
+	public getCommands(): string[] {
+		return [...this.commands]
 	}
 
 	/**
 	 * Getter for the environment this job must run under.
 	 */
 	public getEnvironment(): JobEnv {
-		return this.environment
+		return { dockerImage: this.environment.dockerImage }
```
rvighne, line 122: Why? This will make it so you have to update this method every time `JobEnv` changes.

src/Job/NormalJob.ts
```diff
@@ -14,64 +15,110 @@ const DEFAULT_ENV: JobEnv = Object.freeze({
...
+	constructor(options: NormalJobOptions) {
+		// Make copies so the caller can't directly access prerequisites. Encapsulation!
+		this.prerequisiteJobs = new Set(options.prerequisiteJobs) // We don't need a deep copy because Jobs are immutable.
+		this.prerequisiteFiles = new Set(options.prerequisiteFiles)
+		this.commands = options.commands.slice()
+		this.target = options.target
+		this.environment = options.environment
+			? {
+					dockerImage: options.environment.dockerImage,
+			  }
```
rvighne, line 46: You can automate this object cloning. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/Controller/MakefileTraceToJobTree.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/**
+ * Creates a DAG tree of Jobs from a trace.
+ *
+ * @param trace - the output from a call to 'make --trace --dry-run'.
+ * @returns the tree's root jobs.
+ */
+export function makefileTraceToJobTree(trace: string): Set<Job> {
+	const traceLines = trace.split('\n')
```
rvighne, line 206: If make on windows prints with `\r\n`, this would be a problem.

src/Controller/MakefileTraceToJobTree.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/**
+ * Filters an array of trace-lines.
+ * Removes lines that are unnecessary to the DAG constructor.
+ *
+ * @param traceLines - an array of strings, each of which is a line from a trace.
+ * @returns the filtered array of strings.
+ */
+function filterTraceLines(traceLines: string[]): string[] {
```
rvighne, line 170: How do you know this function is exhaustive? I.e. that there isn't some rarely used output from make that will break this if it appears? If possible, use a whitelist approach instead of a blacklist (but idk anything about the format, so that might not be possible).

src/Controller/MakefileTraceToJobTree.ts
```diff
@@ -0,0 +1,215 @@
...
+	const targetToJAD = new Map<string, JobAndDependency>()
+
+	// Construct DAG.
+	for (const { targetLine, commands } of targetLinesWithCommands) {
+		// Extract target and prerequisites from target line.
+		const { target, prerequisiteJobTargets } = extractInfoFromTargetLine(
+			targetLine,
+		)
+
+		// Find all prerequisite jobs. If no Job has some target, then the target is (hopefully) a preexisting file and we can ignore it. Also marks prerequisite Jobs as having dependents.
```
rvighne, line 120: nitpick: split this over multiple lines, I don't like to scroll horizontally :)

src/Controller/MakefileTraceToJobTree.ts
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Reads a line from a trace containing a target.
+ * Extracts the target and any referenced prerequisites.
+ *
+ * The target line must match one of two recognized Regular Expressions.
+ *
+ * @param targetLine - a line from a trace containing a target.
+ * @returns the target and prerequisites.
+ */
+function extractInfoFromTargetLine(
```
rvighne, line 43: Possibly a weird suggestion, but could you move this function directly above the one that calls it (constructDAGFromTargetsAndCommands)? That makes it easier to read and know what is related.

rvighne (changes requested):
Solid work ✨✨

This is amazingly thorough. I reviewed as much as I could without knowing much about makefiles.

src/Job/NormalJob.ts
Rohanator9000: How would you do it?

src/Controller/MakefileTraceToJobTree.ts
Rohanator9000: How do I check whether a line is a command or not?

Maybe I could throw an error if there are any lines starting with `make: ` remaining at the end of the function, but I'm not sure that a command couldn't start with that.

src/Controller/MakefileToJobTree.ts
Rohanator9000: What should I do with the fields? Just leave them there in case someone wants to use them in the future?

src/Controller/MakefileTraceToJobTree.ts
yashlala: It should be fine. I took a look at GNU Make's source code (a relevant output snippet is attached below), and it seems like we're testing for all possibilities. 


```
  /* Trace the build.
     Use message here so that changes to working directories are logged.  */
  if (trace_flag)
    {
      char *newer = allocated_variable_expand_for_file ("$?", c->file);
      const char *nm;

      if (! cmds->fileinfo.filenm)
        nm = _("<builtin>");
      else
        {
          char *n = alloca (strlen (cmds->fileinfo.filenm) + 1 + 11 + 1);
          sprintf (n, "%s:%lu", cmds->fileinfo.filenm, cmds->fileinfo.lineno);
          nm = n;
        }

      if (newer[0] == '\0')
        OSS (message, 0,
             _("%s: target '%s' does not exist"), nm, c->file->name);
      else
        OSSS (message, 0,
              _("%s: update target '%s' due to: %s"), nm, c->file->name, newer);

      free (newer);
    }
```

Rohanator9000: Where do the `make: Entering directory` strings come from?

src/Controller/MakefileToJobTree.ts
rvighne: I believe they will get printed out whenever the error is console.logged or similar. You don't need to do anything with the fields aside from storing them.

src/Job/Job.ts
```diff
@@ -13,30 +13,43 @@ export interface JobEnv {
...
 	 *
-	 * This method may be removed in the future because Jobs may no longer have names.
+	 * The target is the resulting file.
 	 *
-	 * @experimental
-	 * @returns This job's name.
+	 * @returns This target.
 	 */
-	getName(): string
+	getTarget(): string
```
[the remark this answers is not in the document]
Rohanator9000: Maybe I could set the original method `getName()` as `@deprecated` and add `getTarget()` to return the same thing. I want to add the `getTarget()` somewhere because someone running a Job and trying to find its target shouldn't have to know that its name is actually its target.

Thoughts?

src/Job/NormalJob.ts
rvighne: They have an example on that page under <q>Cloning an object</q>. Something like
```ts
this.environment = Object.assign({}, options.environment)
```
The benefit of this is that you don't need to know the properties of JobEnv

Rohanator9000: `Object.assign` only shallow copies. Meaning if `JobEnv` gets a new property that's an object, someone could manipulate that and change a Job's internal state. I specified `dockerImage` explicitly so that any changes to `JobEnv` would have to be directly addressed here.

src/Job/Job.ts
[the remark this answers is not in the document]
rvighne: Sure that's a good way to do it. Backwards compatible while still encouraging the new method.

src/Controller/MakefileToJobTree.ts
Rohanator9000: I switched from `exec` to `execFile()` which escapes its arguments.

Also, this is a relatively harmless security flaw since we're only running this on the local machine. If a user really wanted to input a malicious command, they could just do it themselves in terminal.

src/Job/NormalJob.ts
rvighne: Ah I see. If you want to get around that you can do a recursive object clone; there's plenty of examples of that out there.

> so that any changes to JobEnv would have to be directly addressed here.

I don't think this should be the case. `Job` isn't responsible for executing the job, only for tracking data about the job. So it's a bit unexpected if someone wants to change JobEnv and has to change it here as well. Relying on the internal details more than it needs to.

rvighne (approved):

PR #76, opened 2021-03-03
SSH-based client

### LINKED ISSUE #12: Client sends jobs to daemons
## User Story
As a developer, I would like the recipes and environment information to be sent to the daemon so that jobs can be run in parallel.

## Detailed Description
Given that the job sorting module (#50) has linearized the DAG of jobs along with an API to find jobs that can run in parallel, the client should send those jobs to the daemon using SSH.

## Tasks
- [ ] Establish an SSH tunnel to the daemon (accept all host keys, as authentication is not a primary concern for now).
- [ ] Send an execute request over the SSH channel and and asynchronously wait for the result.
- [ ] Record success or failure to the job sorting module, and to the user when the job sorter declares completion.
- [ ] Write unit tests.

## Acceptance Criteria
- [ ] Given a new daemon is discovered somehow, then the client opens a new SSH connection to it without prompting.
- [ ] Given a job is ready but not started and a known daemon is idle, then the job is sent to the daemon.
- [ ] Given all sub-tasks are finished, the user is notified.

## Estimated points
1

rvighne:
I significantly refactored the Client into a `GenericClient` so that it can handle arbitrary connection methods. Then, I added a `SSHConnection` strategy that implements the SSH client and made it the default.

It has been tested (and works as expected) end-to-end against the SSH daemon, including
- job ordering + failure recovery for both network (recoverable) and job (propagate) errors.
- triggering jobs on Docker.
- network discovery (this worked without modification same as it did with HTTP).
- basic unit tests.

Closes #12


bin/controller.ts
```diff
@@ -2,37 +2,36 @@
...
-const client: Client = new Http2Client(new HeapJobOrderer([job5]))
-client.on('progress', (finishedJob) => {
-	console.log(`${finishedJob.getName()} completed\n`)
+const client: Client = new GenericClient(
+	createSSHConnection,
+	process,
+	new HeapJobOrderer([job5]),
+)
+client.on('error', console.error)
+client.on('progress', (job, result) => console.log(job, result.status))
```
yashlala, line 29: It seems that job's `toString()` method isn't defined. 
If I understand JS string representations correctly, we'll only print out something like `[NormalJob]`, which doesn't really give any information to the user. Is this a problem, or am I wrong?

src/Job/NormalJob.ts
```diff
@@ -8,13 +18,16 @@ export class NormalJob implements Job {
 
 	/**
 	 * @param name - The job's name. Must be unique between jobs in the same dependency graph.
-	 * @param prerequisites - An optional set containing all of this job's prerequisites. Defaults to no prerequisites.
+	 * @param prerequisites - A set containing all of this job's prerequisites. Defaults to no prerequisites.
+	 * @param environment - A description of the job's runtime environment. For forwards compatibility, always provide this parameter. Defaults to a Docker image containing GNU Make.
```
yashlala, line 22:
```suggestion
	 * @param environment - A description of the job's runtime environment. For forwards compatibility, always provide this parameter. Defaults to the name of a Docker image containing GNU Make.
```

test/unit/SSHConnection.test.ts
```diff
@@ -0,0 +1,7 @@
+describe('SSHConnection', () => {
```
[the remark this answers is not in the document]
rvighne: Low on time. There's no other functionality needed to implement them but I just wanted to get the code out quickly.

src/Client/Http2Connection.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Create a new connection to a daemon using HTTP/2.
+ *
+ * @param host - Hostname or IP address of the daemon.
+ * @param port - The daemon's port number on the host.
+ * @returns A connection to the daemon, or the underlying HTTP/2 error.
+ *
+ * @deprecated Use an SSH connection (with an SSH daemon) instead.
+ */
+export const createHttp2Connection: ConnectionFactory = async (host, port) => {
```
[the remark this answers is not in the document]
rvighne: No reason to delete it just yet. It could be useful for writing quick tests or manual testing.

bin/controller.ts
rvighne: `console.log` is special. It prints out the whole object nicely with colors. Though, in this case, `NormalJob` does have a `toString` so it'll print that.

src/Job/NormalJob.ts
rvighne: I want to keep this language generic (just like I made a `JobEnv` interface as opposed to a string containing the image name). There may be other details aside from the image name that we need in the future.

test/unit/SSHConnection.test.ts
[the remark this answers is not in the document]
rvighne: Also, this made me realize that they should really be integration tests, not unit. I split out the pure-logic part of it, wrote unit tests for that, and moved this into `test/integ` instead of `test/unit`.

yashlala (approved):

pjot726 (approved):
Finished reviewing code, I think I have a good grasp on things

PR #79, opened 2021-03-06
Added Images folder for the user-manual

erlstanleylee:


rvighne:
Why include the images with the source code? Can the wiki not embed images?

erlstanleylee:
Markdown requires that the image has a link.

rvighne (changes requested):
Looks like the way to add images to the wiki is to clone it locally and commit the image there: https://stackoverflow.com/questions/10045517/embedding-images-inside-a-github-wiki-gollum-repository

The clone link for our wiki appears to be https://github.com/cs130-w21/1.wiki.git

erlstanleylee:
Thank you for the resource. It works. I suppose we can then delete the branch and the pull request.

PR #81, opened 2021-03-07
Client side file transfer code

pjot726:
This code handles the file transfer across SSH from the client side.


src/Client/SSHConnection.ts
```diff
@@ -40,6 +45,30 @@ export const createSSHConnection: ConnectionFactory = async (host, port) => {
...
+			// TBD  peipe tar.c to stdin of exec
+			// We send a put request to the daemon
+			const putPayload = JSON.stringify(jobToPushInputs(job))
+			const pushInputStream = await promisify(conn.exec.bind(conn))(putPayload)
+			// We generate list of jobs to be tarred in the stream.
+			const arrayOfJobs: Array<Job> = job.getDeepPrerequisitesIterable()
+			const fileList: Array<string> = []
+			for (const subJob of arrayOfJobs) {
+				fileList.push(subJob.getName())
+			}
```
Rohanator9000, line 58: Should you also be calling `getPrerequisiteFilesIterable()` here? The `Job` might need files that aren't targets.

src/Job/Job.ts
```diff
@@ -48,6 +48,13 @@ export interface Job {
 	 */
 	getPrerequisiteFilesIterable(): Iterable<string>
 
+	/**
+	 * Gets this Job's dependencies (its prerequisites, recursively)
+	 *
+	 * @returns an array containing a deep scan of the Job's prerequisites.
+	 */
+	getDeepPrerequisitesIterable(): Array<Job>
```
Rohanator9000, line 56: Why not make this return a `Set`? It seems like order doesn't matter.

src/Job/NormalJob.ts
```diff
@@ -79,7 +79,26 @@ export class NormalJob implements Job {
...
 
 	/**
-	 * Returns the number of prerequisite Jobs using Set's native size property.
+	 * Gets this Job's dependencies (its prerequisites, recursively)
+	 *
+	 * @returns a iterable containing a deep scan of the Job's prerequisites.
+	 */
+	public getDeepPrerequisitesIterable(): Array<Job> {
+		let childrenPrereqs: Array<Job> = []
+		if (this.getNumPrerequisiteJobs() !== 0) {
```
Rohanator9000, line 88: You don't need this `if` statement.

src/Job/NormalJob.ts
```diff
@@ -79,7 +79,26 @@ export class NormalJob implements Job {
 	}
 
 	/**
-	 * Returns the number of prerequisite Jobs using Set's native size property.
+	 * Gets this Job's dependencies (its prerequisites, recursively)
+	 *
+	 * @returns a iterable containing a deep scan of the Job's prerequisites.
+	 */
+	public getDeepPrerequisitesIterable(): Array<Job> {
+		let childrenPrereqs: Array<Job> = []
```
Rohanator9000, line 87: I find the name `childrenPrereqs` kinda misleading since it holds more than just the children prerequisites by the end of the function.

src/Job/NormalJob.ts
```diff
@@ -79,7 +79,26 @@ export class NormalJob implements Job {
...
+		if (this.getNumPrerequisiteJobs() !== 0) {
+			for (const prereqJob of this.getPrerequisiteJobsIterable()) {
+				childrenPrereqs = childrenPrereqs.concat(
+					prereqJob.getDeepPrerequisitesIterable(),
+				)
+			}
+		}
+		childrenPrereqs.concat(Array.from(this.getPrerequisiteJobsIterable()))
```
Rohanator9000, line 95: Why not just do this during the prior `for` loop? `Array.from` is O(n).

test/unit/NormalJob.test.ts
```diff
@@ -116,6 +116,38 @@ describe('NormalJob', () => {
...
+		const job2 = new NormalJob({ target: '2', commands: [], environment })
+		const job3 = new NormalJob({ target: '3', commands: [], environment })
+		const job4 = new NormalJob({ target: '4', commands: [], environment })
+		const job5 = new NormalJob({ target: '5', commands: [], environment })
+		const job6 = new NormalJob({ target: '6', commands: [], environment })
+		const prerequisites1 = new Set<Job>([job1, job2, job3])
+		const prerequisites2 = new Set<Job>([job4, job5, job6])
+		const job7 = new NormalJob({
+			target: '7',
+			prerequisiteJobs: prerequisites1,
```
Rohanator9000, line 131: Since you don't use `prerequisites1` anywhere else, and the name isn't particularly indicative, I think it'd be more readable if you just replaced it with the `Set`. Same for `prerequisites2` and `prerequisites3`.

src/Job/Job.ts
Rohanator9000: And if you're going to make it return a `Set`, it'll probably be easiest to just return an `Iterable` like the other functions.

Rohanator9000 (changes requested):

src/Client/SSHConnection.ts
pjot726: I asked @rvighne for clarification regarding this. It is my understanding that the job name is all we need.

src/Job/Job.ts
pjot726: I think order matters, because the way it is right now, pre-req jobs occur before the jobs that require them in the array. If this behavior has no function when it coming to tarball generation later on, I think we can change it

src/Client/SSHConnection.ts
rvighne: Yeah I was also under the impression that job name = target name in the Makefile = file name on disk

test/unit/NormalJob.test.ts
pjot726: Fixed

src/Client/SSHConnection.ts
Rohanator9000: Yes, but what about a preexisting file `a.c` that needs to be there from the start, instead of being built during the process?

rvighne: I see, yeah that's important. @pjot726 can you make sure you take care of this situation?

As I understand it, if we have
```makefile
intermediate.txt: input.txt
    ...

output.txt: intermediate.txt
    ...
```
there will be `Job`s created for `output.txt` and `intermediate.txt`, but not `input.txt`, and to handle that we need the function Rohan recommended.

src/Job/NormalJob.ts
pjot726: Fixed

src/Client/SSHConnection.ts
pjot726: Latest push should address this issue

erlstanleylee (approved):

alexzma (approved):

PR #82, opened 2021-03-07
Temp volume tests

alexzma:
These are tests for the temp volume creation.

### DIFF
--- a/test/unit/TempVolume.test.ts
+++ b/test/unit/TempVolume.test.ts
@@ -4,7 +4,7 @@ import {
 	safeResolve,
 	createTempDir,
 	destroyTempDir,
-	withTempDir
+	withTempDir,
 } from '../../src/Daemon/TempVolume'
 
 describe('safeResolve', () => {
@@ -52,6 +52,7 @@ describe('createTempDir', () => {
 	})
 	it('ensures the tempdir cannot be accessed by others', async () => {
 		const stat = await fs.stat(tempdir)
+		// eslint-disable-next-line no-bitwise
 		expect(stat.mode & 511).toBe(448)
 	})
 })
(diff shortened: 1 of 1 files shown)


test/unit/TempVolume.test.ts
```diff
@@ -1,25 +1,125 @@
...
+		expect(path2).toEqual('/usr/bin/ls')
+	})
+	it('rejects relative paths that escape the root', () => {
+		const resolved = safeResolve('../', 'ls')
+		expect(resolved).toBeUndefined()
+	})
+	it('rejects absolute paths that escape the root', () => {
+		const resolved = safeResolve('/../', 'ls')
+		expect(resolved).toBeUndefined()
+	})
```
rvighne, line 27: `safeResolve` needs the first parameter (`root`) to be an absolute path, but you're testing relative paths here. The cases we actually want to prevent are like this:
```ts
safeResolve('/bin/safe', '../risky')
safeResolve('/bin/safe', '/sbin/su')
```

test/unit/TempVolume.test.ts
```diff
@@ -1,25 +1,125 @@
...
+		expect(existsSync(filepath)).toBeTruthy()
+	})
+	it('ensures the tempdir cannot be accessed by others', async () => {
+		const stat = await fs.stat(tempdir)
+
+		// 511 is octal for 777, 448 is octal for 700
+		// this is testing that the permissions are only for the user
+		// see https://man7.org/linux/man-pages/man2/chmod.2.html for more details
+		// eslint-disable-next-line no-bitwise
+		expect(stat.mode & 511).toBe(448)
```
rvighne, line 60: 🤩 very thorough

test/unit/TempVolume.test.ts
```diff
@@ -1,25 +1,125 @@
...
+		expect(existsSync(directory)).toBeFalsy()
+	})
+	it('removes the tempdir even when the action throws', async () => {
+		const action = async (root: string): Promise<string> => {
+			throw new Error(root)
+			return root
+		}
+
+		await withTempDir(action).catch((error: Error) => {
+			expect(existsSync(error.message)).toBeFalsy()
```
rvighne, line 111: Interesting, this is a lot more concise than what I had in mind.

rvighne (changes requested):
Wow you really went all-out on writing these tests 😆. Thank you!

rvighne (approved):

yashlala (approved):

PR #83, opened 2021-03-08
Fix 'npm run cli' bug

Rohanator9000:
- before it was throwing an error
- now it prints help and then exits with code 1

### DIFF
--- a/bin/controller.ts
+++ b/bin/controller.ts
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 import bonjour from 'bonjour'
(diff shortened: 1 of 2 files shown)


yashlala:
I'm so confused. Why does the imported `yargs` object have the same state as the one I modified in the CLI controller module? How does it now what message to print? 

I've confirmed this works. I'm not sure why, but we don't have the time. Mergy mergy.

yashlala (approved):

rvighne (approved):

